PR #1, opened 2015-03-31
Add a Vagrantfile for local testing

dcarley:
Which mimics the four machines currently in `inventory`. This will be useful
for offline testing or working in parallel to other people.

I found that I had to write the inventory file dynamically because I
couldn't assign the `internal_ip` and `external_ip` variables when using
Vagrant's `groups` support. I think I've also mixed up the terminology for
"hosts" and "roles".

I'm using the PuppetLabs Ubuntu 14.04 box because it's pretty minimal and
supports both VirtualBox and VMware (which the official Ubuntu box doesn't).

There are some ordering issues to work out. It takes several runs on all
machines for them to complete successfully. This might account for the
non-numerical order of the original `inventory` file. We might be able to
help this by breaking each component out to it's own machine and ensuring
that databases happen first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+inventory.vagrant
+.vagrant/
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -58,11 +58,14 @@ Vagrant.configure(2) do |config|
     v.vmx["memsize"] = MEMORY
   end
 
-  config.vm.provision "ansible" do |ansible|
-    ansible.raw_arguments = "--private-key=~/.vagrant.d/insecure_private_key"
-    ansible.verbose = "vv"
-    ansible.playbook = "site.yml"
-    ansible.extra_vars = "globals.yml"
-    ansible.inventory_path = INVENTORY_FILE
+  config.vm.define HOSTS.last[:name] do |host|
+	host.vm.provision :ansible do |ansible|
+ 	  ansible.raw_arguments = "--private-key=~/.vagrant.d/insecure_private_key"
+	  ansible.verbose = "vv"
+	  ansible.playbook = "site.yml"
+	  ansible.extra_vars = "globals.yml"
+	  ansible.inventory_path = INVENTORY_FILE
+	  ansible.limit = "all"
+    end
   end
 end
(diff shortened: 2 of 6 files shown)


dcarley:
Blocked by a problem with the [`generate tsuru token`](https://github.com/alphagov/tsuru-ansible/blob/9b745014ea8662764ea2a0735587064fdf4f79f7/site.yml#L34-L37) task (which uses `delegate_to`) failing in Vagrant because it's not able to SSH from tsuru-i2 to tsuru-i1:

```
TASK: [generate tsuru token] **************************************************
fatal: [tsuru-i2 -> 172.18.10.11] => SSH Error: Permission denied (publickey,password).
    while connecting to 172.18.10.11:22
It is sometimes useful to re-run the command using -vvvv, which prints SSH debug output to help diagnose the issue.
```

annashipman:
OK, that took a while. Here's what happened to me when I tested it; it timed out on each box in turn:

```
<172.18.10.14> REMOTE_MODULE setup
fatal: [tsuru-i4] => SSH Error: ssh: connect to host 172.18.10.14 port 22: Operation timed out
    while connecting to 172.18.10.14:22
It is sometimes useful to re-run the command using -vvvv, which prints SSH debug output to help diagnose the issue.

TASK: [docker_server | Add docker apt key.] ***********************************
FATAL: no hosts matched or all hosts have already failed -- aborting
```

It did that on 1, then I reran and it timed out on 2, then  3, then 4, then finally it's finished and fine and I have four machines running.

Running it four times seems a little arduous though. Any ideas why this is?

mikkoc:
Hmm, very strange..
I don't think it helps, but can you try reverting my commit?
https://github.com/alphagov/tsuru-ansible/commit/c2d02f4a1bf5c3a81c0eeee3bc2405ea16e9fc81
just in case...

annashipman:
Yep will try that. Hold on...

annashipman:
No, makes no difference, still timing out. Hmm.

annashipman:
Maybe someone else could test this? See if it's just me?

Grindizer:
I changed the way vagrant provision the vms, we only do it once after all instances are created. 
I also rebased this branch to include last fix from the master (was necessary to have a working branch).

PR #4, opened 2015-04-01
Add support for multiple routers

mikkoc:
By removing 'xip.io' we are using the CNAME which we previously created with Terraform.
The CNAME is pointing to the Load balancer that sits in front of the router instances.


mikkoc:
This breaks deployments that rely on xip.io, like a local Vagrant deployment.
We need to think about an alternative solution.

annashipman:
@mikkoc What's the story with this PR? It sounds from your second comment that this is not going to work, so shall we just close it?

mikkoc:
It will work, but only with the other PR I am about to make. I can close this one and include it in the big one coming up next.

PR #6, opened 2015-04-07
Reorder playbook

Grindizer:
Separate the playbook per component and reorder them according to the logical dependency.


mikkoc:
Looks good to me.

PR #9, opened 2015-04-09
Configure private instances

mikkoc:
This PR lets us use a bastion host to login to the instances, so we don't need to assign public IPs to them.
It also adds support to multiple routers instances, by using a Load balancer.


Grindizer:
I added the code to be able to have it run with vagrant. 
Basically added .xip.io to all external_ip value in  the inventory.vagrant. It is mainly needed for hipache, but since we agreed that external_ip will hold the external dns name of the service (typically the lb), it is still ok to add .xip.io to all values.

PR #10, opened 2015-04-09
Add librarian-ansible support

actionjack:
https://www.pivotaltracker.com/story/show/91930038

This pull request covers adding [librarian-ansible](https://github.com/bcoe/librarian-ansible) support to our ansible configuration.

We are moving the management of our playbooks to [librarian-ansible](https://github.com/bcoe/librarian-ansible) so that we don't accidentally fork (or extend) our playbooks and we can easily update them.

This pull request does not include migration to a new redis playbook this will come in a follow up pull request.


annashipman:
This PR won't work because we had made some changes to the Redis module we were using (the exact problem we are trying to address by moving to librarian-ansible!) so we are going to close this and make it so that it work.

annashipman:
It wasn't Redis, it was mongodb, so we've made the change in `globals.yml` to make it work. We've also added instructions to the README.

I merged from master rather than rebasing so the history sticks around; this commit will disappear once it's merged to master.

This is now ready for review.

README.md
```diff
@@ -11,6 +11,11 @@ Ansible based project to configure a multi-node tsuru cluster.
 > Note: can be any number of nodes, but every section in the inventory file has to have at least one node. 
 > (it is possible to use a node in more that one section).
```
Grindizer, line 14: Would be nice if we add the librarian dependency with required version to be used here.

annashipman:
FYI, no-one gets a notification when you add a new commit, there are only notifications when you comment, so when a PR is ready to merge after review comments have been addressed, it's worth adding a comment so the original reviewer knows to pick it up again.

Vagrantfile
```diff
@@ -2,7 +2,7 @@
 # vi: set ft=ruby :
 
 INVENTORY_FILE = "inventory.vagrant"
-MEMORY = 384
+MEMORY = 1024
```
dcarley, line 5: I'm going to struggle to bring up 4x 1G VMs on my little laptop.

Vagrantfile
```diff
@@ -41,7 +41,10 @@ end
 generate_inventory(HOSTS, INVENTORY_FILE) if %w{up provision}.include?(ARGV[0])
 
 Vagrant.configure(2) do |config|
-  config.vm.box = "puppetlabs/ubuntu-14.04-64-nocm"
+  config.vm.box = "ubuntu/trusty64"
+  config.ssh.private_key_path = File.expand_path('~/.vagrant.d/insecure_private_key')
+  config.ssh.insert_key = false
```
dcarley, line 46: The commit message doesn't mention why these options were added at the same time. Can you recall?

Vagrantfile
Grindizer, line 5: I have first tested with 512 and still wasn't enough. If there is anyother option I would be happy take it.

Vagrantfile
Grindizer, line 46: if not the insecure_key will not be inserted in the instance and vagrant ssh will ask for a password.
I supposed that unlike the puppetlab box the ubuntu one doesn't include this feature automatically.

PR #13, opened 2015-04-13
Added postgresql server to automatic provisioning

jimconner:
There are actually four subcomponents to this pull request.

1.) Add support for ansible-vault for storing secure information (i.e. passwords).
2.) Move existing passwords from globals.yml to the vault.
3.) Add an extra VM to the Vagrant file for hosting postgres and use the vault.
4.) Ansible provisioning for postgresql server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,4 @@ ssh-add <the-public-ssh-key-file>
 ```{r, engine='bash'}
 ansible-playbook -i inventory site.yml -e "@globals.yml"
 ```
+


Vagrantfile
```diff
@@ -68,6 +70,7 @@ Vagrant.configure(2) do |config|
 	  ansible.playbook = "site.yml"
 	  ansible.extra_vars = "globals.yml"
 	  ansible.inventory_path = INVENTORY_FILE
+	  ansible.vault_password_file = "~/.vault_pass.txt"
```
Grindizer, line 73: "vagrant up" failed because it doesn't find this file, I guess this is specific to the local setup. 
may be better to use "--ask-pass" here instead since it will be more generic ?
(Or at least describe how this file would looks like and that it has to exist in the Readme)

globals.yml
```diff
@@ -22,5 +22,3 @@ mongodb_conf_bind_ip: 0.0.0.0
 mongodb_host: "{{ hostvars[groups['mongodb'][0]].internal_ip }}"
 mongodb_port: 27017
 
-admin_user: administrator@gds.tsuru.gov
-admin_password: admin123
```
Grindizer, line 24: How about just commenting those vars here and also add the ones defined in the secure file ?
The problem I see here is that once the vars secured with vault we have no idea about what are the vars names. 
It makes things hard to debug, not to mention that the project becomes almost impossible to reuse.

PR #14, opened 2015-04-16
91206714/resiliant docker registry

Grindizer:
Add support to setup S3 and GCS as storage backend for the docker-registry. 
When using this patch don't forget to update the docker-registry role with librarian, since it works with 1.0.4 (instead of 1.0.3).

1.0.4 adds support for GCS backend (https://github.com/codingbunch/ansible-docker-registry/commit/432e989ca0365500f40eacace3a66a8acdfe3c14)

### DIFF
--- a/Ansiblefile
+++ b/Ansiblefile
@@ -3,7 +3,7 @@
 
 site "https://galaxy.ansible.com/api/v1"
 
-role "aalda.docker-registry", "1.0.3"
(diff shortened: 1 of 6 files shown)


jimconner:
I've now tested this on aws and it all deploys OK once I worked out which variables needed to be passed on the command line ( docker_registry_port ) - Perhaps we should update the README.md with the full ansible-playbook command which we now need to use.

PR #16, opened 2015-04-17
Capture dependencies with pip and bundler

dcarley:
For Python and Ruby respectively. This allows people to install the
dependencies programmatically and reduces the risk that our documentation is
out of step with the code (although pip doesn't enforce this as well as
bundler).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,16 @@
 
 Ansible based project to configure a multi-node tsuru cluster.
 
-### Requirement:
-* ansible 1.9 or higher.
-* librarian-ansible 1.0.6 or higher
+### Requirements:
+
+* Python things (you may wish to use [virtualenv](https://virtualenv.pypa.io/en/latest/)):
+```
+pip install -r requirements.txt
+```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2,9 +2,16 @@
...
 Ansible based project to configure a multi-node tsuru cluster.
 
-### Requirement:
-* ansible 1.4 or higher.
-* librarian-ansible 1.0.6 or higher
+### Requirements:
+
+* Python things (you may wish to use [virtualenv](https://virtualenv.pypa.io/en/latest/)):
+```
+pip install -r requirements.txt
```
bradwright: If you use `-Ur` it'll also make sure we [upgrade to the latest specified package versions](https://pip.pypa.io/en/latest/reference/pip_install.html#cmdoption-U)

dcarley: Added, thad.

dcarley:
Rebased to fix merge conflict.

mikkoc:
Looks good to me

PR #22, opened 2015-04-27
Remove external_ip from postgres node

dcarley:
This is not used so we don't need to maintain/update it.


annashipman:
:+1: LGTM

PR #25, opened 2015-04-27
Updated IP of postgres instance that has changed

lucorsel:
Updated IP of postgres instance that has changed and tested. Please merge to master.


mikkoc:
LGTM

PR #26, opened 2015-04-27
[#91206264] Multiple api servers

lucorsel:
Added IPs and reference to two newly generated API servers that replace the legacy one.

The legacy API server (.22) remains running because it mongo and redis have been historically installed there.


dcarley:
For future reference, the commit 07a132fc70f56f5a798262dc82eb9d96a1f4b885 doesn't provide much value by itself. It could have either been squashed into the original commit or the commit message should have described the behaviour before and after the change.

PR #29, opened 2015-04-28
Remove extraneous entries from AWS inventory

dcarley:
@jimconner and I talked about why we have these entries duplicated at the
top of the inventory file and within each role. Some are only listed within
roles. We couldn't find a reason why and removing them doesn't appear to
make any difference when running `ansible-playbook` against a new cluster.
So remove them and make it easier to maintain.


annashipman:
I would probably rather have all of the things that change declared in a block at the top (plus the ones that aren't currently there) and then variables used within the roles, is that possible?

dcarley:
I have no idea, sorry.

We probably shouldn't spend too much time on it though if it's going to be replaced by a dynamic inventory.

mikkoc:
I agree with @jimconner and @dcarley on this one. It does not make sense to have them duplicated :+1: 
@annashipman we are indeed using variables within the roles. i.e.: `groups['mongodb'][0]].internal_ip`

annashipman:
@mikkoc yep, definitely agree that we should avoid duplication, but I was suggesting we remove duplication in the other direction, i.e. put the IP addresses in at the top rather than scattered throughout.

However, it's not a huge document and @dcarley makes a good point that it's going to be replaced anyway, so it's not worth spending any time on that. Will merge.

PR #30, opened 2015-04-28
[#91206264] Added delegation so that admin user team changes are run from mongo

lucorsel:
API and mongo are no longer supposed to be running on the same instance,
mongo-clients is not installed on the api instances and the mongo server
needs to be address to add necessary modifications to admin user teams.

The operation on mongo is nog delegated to the appropriate mongo
instance (the first mongo instance that is present)

### DIFF
--- a/roles-static/tsuru_api/tasks/main.yml
+++ b/roles-static/tsuru_api/tasks/main.yml
@@ -30,19 +30,3 @@
 
 - name: Wait for tsuru-server to be up
   wait_for: port={{api_port}} timeout=60
-
-# TODO: handle the case where the user already exists
-- name: Add admin user team
-  shell: >
-    mongo tsuru --eval 'db.teams.update({_id: "admin"}, {_id: "admin"}, {upsert: true})';
-    mongo tsuru --eval "db.teams.update({_id: 'admin'}, {\$addToSet: {users: '{{admin_user}}'}})";
-  delegate_to: "{{ groups['mongodb'][0] }}"
-  
-- name: Add admin user.
-  shell: >
-    curl -sS -XPOST -d"{\"email\":\"{{admin_user}}\",\"password\":\"{{admin_password}}\"}" "http://127.0.0.1:{{api_port}}/users"
-
-- name: Login with the admin user.
-  shell: >
-    token=$(curl -sS -XPOST -d"{\"password\":\"{{admin_password}}\"}" "http://127.0.0.1:{{api_port}}/users/{{admin_user}}/tokens" | jq -r .token);
-    echo "${token}" > ~/.tsuru_token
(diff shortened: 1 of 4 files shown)


dcarley:
It feels like these tasks should probably be moved to `site.yml` to match the delegation we're doing for Gandalf and because it's not strictly related to the setup of a Tsuru API machine (more to do with the orchestration of the entire platform, if that makes sense).

roles-static/tsuru_api/tasks/main.yml
```diff
@@ -32,13 +32,17 @@
   wait_for: port={{api_port}} timeout=60
 
 # TODO: handle the case where the user already exists
-- name: Add admin user.
+- name: Add admin user team
   shell: >
     mongo tsuru --eval 'db.teams.update({_id: "admin"}, {_id: "admin"}, {upsert: true})';
     mongo tsuru --eval "db.teams.update({_id: 'admin'}, {\$addToSet: {users: '{{admin_user}}'}})";
```
dcarley: If we only get the exit code of the last command, should we split these into separate tasks to surface errors better?

lucorsel: It seems to me that apart from the exit code we see the output of both commands in ansible, in this case enough to understand what might have gone wrong. Here an execution output, result of both commands looks reasonably readable:

changed: [10.128.1.200 -> 10.128.1.22] => {"changed": true, "cmd": "mongo tsuru --eval 'db.teams.update({_id: \"admin\"}, {_id: \"admin\"}, {upsert: true})'; mongo tsuru --eval \"db.teams.update({_id: 'admin'}, {\$addToSet: {users: 'administrator@gds.tsuru.gov'}})\";", "delta": "0:00:00.095792", "end": "2015-04-28 15:15:46.171138", "rc": 0, "start": "2015-04-28 15:15:46.075346", "stderr": "", "stdout": "MongoDB shell version: 2.6.9\nconnecting to: tsuru\nWriteResult({ \"nMatched\" : 1, \"nUpserted\" : 0, \"nModified\" : 1 })\nMongoDB shell version: 2.6.9\nconnecting to: tsuru\nWriteResult({ \"nMatched\" : 1, \"nUpserted\" : 0, \"nModified\" : 1 })", "warnings": []}

lucorsel:
In site.yaml, shall I put these user-creation related commands still under tsuru_api with the delegation to the mongo server or where otherwise?

lucorsel:
Reimplemented the same function with the same delegation principle into site.yaml, as suggested and then tested successfully.

Please check last commit , evaluate and eventually merge or comment. Thanks.

lucorsel:
Please note: I left the previous commit on purpose into this request, to leave a reasonably readable  historic trail showing the difference among the first simple change and the further move of the same into site.yaml.

If so desired I can rebase and let the first commit disappear.

annashipman:
Thanks for the changes. I agree that it makes more sense to have the delegation in `site.yml` as that seems to be where we're pulling the whole platform together. Will be useful to be able to see the separate exit codes as well.

lucorsel:
You are welcome. Both move to site.yams and separated commands are implemented in that last commit, I thought it was useful to specify this into the PR as well as in the commit message, for clarity.

PR #31, opened 2015-04-29
Workaround hipache bug when IPv6 is disabled

dcarley:
Unlike AWS our GCE instances don't have IPv6 support enabled by default
(there are no link-local addresses). A bug in Hipache causes workers to die
when they attempt to listen on `::1`. The following errors gets repeated:

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: bind EADDRNOTAVAIL
    at errnoException (net.js:904:11)
    at net.js:1072:30
    at Object.2345:2 (cluster.js:592:5)
    at handleResponse (cluster.js:171:41)
    at respond (cluster.js:192:5)
    at handleMessage (cluster.js:202:5)
    at process.EventEmitter.emit (events.js:117:20)
    at handleMessage (child_process.js:318:10)
    at Pipe.channel.onread (child_process.js:345:11)
29 Apr 09:44:22 - Worker died (pid: 16453, suicide: false, exitcode: 8). Spawning a new one.
```

Fix this by explicitly not binding to any v6 addresses when Ansible's fact
hasn't registered any.

### DIFF
--- a/roles-static/hipache/templates/hipache.conf.j2
+++ b/roles-static/hipache/templates/hipache.conf.j2
@@ -2,6 +2,9 @@
     "server": {
         "debug": true,
         "accessLog": "/var/log/hipache/access.log",
+{% if ansible_all_ipv6_addresses == [] %}
+        "address6": [],
+{% endif %}
         "port": 80,
         "workers": 5,
         "maxSockets": 100,


mikkoc:
Tested on AWS, it works fine :+1: 
Although, I would expect `"address6": [],` to show up in the `/etc/hipache.conf` but it doesn't.
Nonetheless, it binds to the ipv6 anyway:

```
tcp6       0      0 ::1:80                  :::*                    LISTEN      15354/node 
```

jimconner:
The "address6": [] doesn't show up on AWS because the AWS instances have IPV6 support, meaning that hipache can bind to '::1' OK - It is for GCE that we need the 'address6' line added to the config.

dcarley:
Thanks @mikkoc !

It shouldn't add the line in AWS because the machines have v6 addresses and Hipache's default of `"address6": [ "::1" ],` works correctly.

mikkoc:
Oh right, silly me, I totally misread that conditional!

PR #33, opened 2015-04-30
Moved globals.yml into group_vars/all

jimconner:
The group_vars tree is automatically included by ansible - placing the file in here means that we don't need to specify it on the command line with every ansible-playbook run.

No changes have been made to any file contents - this is just a move.


mikkoc:
Can you also update the README to reflect the change? Other than that LGTM! :+1:

jimconner:
README.md now updated too

PR #34, opened 2015-04-30
Split site.yml into platform-specific and common

jimconner:
As we are using the same ansible scripts to configure hosts on two different cloud providers, we are starting to need some platform-specific options passed to ansible roles. Specifically, we need the storage_type parameter for docker-registry to be set to 's3', 'gcs' or 'local' for AWS, GCE and Vagrant respectively.
- Renamed site.yml to common.yml
- Created site-aws.yml, site-gce.yml and site-vagrant.yml which include common.yml as their last step.
- Moved docker-registry host definitiion from common.yml into site-aws.yml, site-gce.yml and site-vagrant.yml
- Added storage configuration options to site-aws, site-gce and site-vagrant
- Removed group_vars/docker-registry/storage_backend
- Added S3 access key and secret key to ansible vault
- Updated README.md to reflect changes.


lucorsel:
Added story number for a bit of context. This change is tightly bound to the mentioned story.

site-aws.yml
```diff
@@ -0,0 +1,14 @@
...
+  sudo: yes
+  vars:
+    storage_type: "s3"
+    s3_region: "eu-west-1"
+    s3_bucket: "mcp.registry.storage"
+
+  roles:
+    - aalda.docker-registry
+
+- include: common.yml
```
lucorsel: Just executed a test to create a brand new docker registry, the split config fails because the tasks contained in common.yml need to be executed earlier than the docker registry config.

Problem is solved by putting the "include" line above the "docker-registry" section, please amend.

However I wonder if this current design with site-aws.yml+common.yml will cause other order of execution issues in the future.

actionjack:
:imp:  Looks ok to me but it could have been easier/nicer to review if these were spread into separate commits rather than one big scary one. e.g.
- 1st commit - Created site-aws.yml, site-gce.yml and site-vagrant.yml which include common.yml 
- 2nd commit - Renamed site.yml to common.yml

group_vars/docker-registry/storage_backend
```diff
@@ -1,17 +0,0 @@
...
-gcs_bucket: "gds-docker-registry"
-gcs_storage_path: "/"
-gcs_oauth2: false
-gcs_access_key: "{{ lookup('env', 'GCS_ACCESS_KEY') }}"
-gcs_secret_key: "{{ lookup('env', 'GCS_SECRET_KEY') }}"
-
-# storage_type: "s3"
-s3_region: "eu-west-1"
-s3_bucket: "docker-registry-store"
-s3_storage_path: "/srv/docker"
```
lucorsel, line 15: This variable may need to be added as well if we want to support multiple clusters in the same bucket. By default registry folders are created at top level in the bucket.

group_vars/docker-registry/storage_backend
```diff
@@ -1,17 +0,0 @@
----
-# default storage type is local (file).
-# storage_type: "local"
-
-# storage_type: "gcs"
-gcs_bucket: "gds-docker-registry"
-gcs_storage_path: "/"
-gcs_oauth2: false
-gcs_access_key: "{{ lookup('env', 'GCS_ACCESS_KEY') }}"
-gcs_secret_key: "{{ lookup('env', 'GCS_SECRET_KEY') }}"
```
lucorsel, line 10: Maybe it's better to add google variables to site-gce.yml even if the object storage is not yet implemented on google, so not to lose track of them. Storage_type can stay "local" in there until the functionality is implemented, which will happen very soon.

jimconner, line 10: The gcs storage variables are now added to site-gcs.yml as suggested. This commit has been edited to include them.

PR #35, opened 2015-05-01
[#92502208] Script to store vault password in OS X keychain

dcarley:
It's tedious to type it in each time and it's not good to store it in
plaintext on the filesystem. This is a compromise for people using OS X
until we have something like GPG (and gpg-agent) with individual recipients.


actionjack:
@dcarley Not sure why but this won't merge try rebasing with master.

lucorsel:
There should just be a tiny conflict with readme.md as a consequence of the last merge of Jim's change.

dcarley:
> There should just be a tiny conflict with readme.md as a consequence of the last merge of Jim's change.

Yep. I'll look at this next week.

PR #38, opened 2015-05-07
[#93690844] Use DNS for docker registry

mtekel:
Updating the aws-invertory file to show how we can use DNS record (created and updated by terraform) to point to docker registry.
Use appropriate DNS record `(environment)-docker-registry.tsuru.paas.alphagov.co.uk` in your local inventory files.


actionjack:
LGTM.

PR #39, opened 2015-05-12
Vendor git pre-receive hook

dcarley:
Instead of fetching it from the Internet. This will protect us against
GitHub being down or the script changing and being inconsistent across
environments.

It also means that we're no longer using inline shell script. Which Ansible
was giving us a warning about and we wouldn't have been capturing exit codes
if anything but the last command failed.

The file is taken from this upstream:

https://github.com/tsuru/tsuru/blob/e0699f287dce9a48e3b727a939bb179df5e7e98c/misc/git-hooks/pre-receive.archive-server

This change is a noop when applied to an existing machine. I'm still able to
deploy an app when this is applied to a new Gandalf machine.

### DIFF
--- a/roles-static/gandalf/defaults/main.yml
+++ b/roles-static/gandalf/defaults/main.yml
@@ -2,3 +2,4 @@
 # defaults file for gandalf
 
 gandalf_port: 8080
+gandalf_git_template_dir: /home/git/bare-template
--- a/roles-static/gandalf/files/pre-receive.archive-server
+++ b/roles-static/gandalf/files/pre-receive.archive-server
@@ -1,0 +1,62 @@
+#!/bin/bash -el
+
+# This script uses the archive-server (https://github.com/tsuru/archive-server)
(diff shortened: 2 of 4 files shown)


annashipman:
Good idea. But do we need to worry about changes being made upstream? Can't think of a good way to handle this though...

dcarley:
Potentially, yeah. Only saving graces are:
- puppet-tsuru has done the same thing: https://github.com/tsuru/puppet-tsuru/commit/fd384b07dfb77fa3f908f2e665d6965b564fcdac
- it hasn't changed _that_ much to date: https://github.com/tsuru/tsuru/commits/e0699f287dce9a48e3b727a939bb179df5e7e98c/misc/git-hooks/pre-receive.archive-server

annashipman:
Fair enough. I admit I also had a quick look at puppet-tsuru to see what it had done but couldn't find it :)

As I can't think of a better way of handling the changes upstream and this is a good thing to do, happy to merge.

PR #41, opened 2015-05-14
Add support for GCE & AWS dynamic inventory scripts

jimconner:
Replaces our former static, manually configured inventory files with the dynamic inventory plugins from Ansible (gce.py and ec2.py)

Fix for gce.py to allow it to work on VMs without a public IP address defined.
Removal of host groups, and replacement with regex hostname matching.
globals.yml slimmed down, and platform-specific variables added into platform_aws.yml and platform_gce.yml respectively
group_vars/docker-registry no longer used (we have no groups anymore) - variables from within now included in site-aws and site-gce respectively
Readme updated.


annashipman:
I am on holiday tomorrow so won't be able to review it. It's a large change so I would like to look over it, but I'm very happy for someone else to review/merge tomorrow and I will take a look on Monday; as long as it works like this, any changes I suggest can be dealt with in separate PRs. It may well be perfect of course and no need for suggested changes! :grinning: 

Good work for getting this one finished, it's been a tough one.

jimconner:
GCE does not have the concept of an internal load balancer, so we have to handle the SSL traffic routing slightly differently.

common.yml
```diff
@@ -55,13 +54,13 @@
...
 
-- hosts: gandalf
+- hosts: ~({{ deploy_env }}|tag_Name_{{ deploy_env }})-tsuru-gandalf*
   sudo: yes
   tasks:
     - name: generate tsuru token
       run_once: true
       shell: tsr token
-      delegate_to: "{{ groups['api'][0] }}"
+      delegate_to: "{{ tsuru_api_internal_lb }}"
```
dcarley, line 63: I think we should rename this. It's confusing that we're delegating to a "load balancer" when in fact it's a machine.

mtekel:
Tested on both AWS & GCE - works fine. Also code LGTM...

actionjack:
I tested this on both AWS and GCE and works fine for me.

It was quite a large series of commits that I think we could have rebased this to reduce it into fewer commits with a cohesive narrative.

Other than that great work! :+1:

annashipman:
For completeness, leaving a comment here. We merged this as it was a large PR, but agreed there may be some follow-up work to tidy things up. These are the issues:
1. Inconsistent host-naming between AWS and GCE:
   https://github.com/alphagov/tsuru-ansible/commit/fa60bdaacec6d5d58e60dfb0081ab4939b6d97f8#commitcomment-11209588
2. Rename load balancer machine https://github.com/alphagov/tsuru-ansible/pull/41#discussion_r30394869
3. Get the GCE and AWS scripts from Ansible rather than vendoring them
4. Has this change raised with Ansible? https://github.com/alphagov/tsuru-ansible/commit/665a3c8520a102b9414b18dc6bd913d1bc2d0cdd If so, provide link to issue and we should have a story to chase it up/make change when merged
5. Ditto this one: https://github.com/alphagov/tsuru-ansible/commit/f0b3196624162b00c548e106142ff2d7dc418777
6. In Ansible we have a `secrets.py` https://github.com/alphagov/tsuru-ansible/commit/3668a53e9672af2614a1495a8864e8bfbf5e6cd9 in Terraform an `accounts.json`. Looks like these contain roughly the same information. Should we look at handling this the same way?
7. Lack of internal LB in GCE... I'm not sure whether this is something we should address with our architecture or accept the differences? https://github.com/alphagov/tsuru-ansible/commit/dd45b85573d685f777a509a99cfdbc69ea09223c#commitcomment-11210205

Brett is going to raise stories for these. If you have any comments on any of them please add here or to Pivotal.

PR #43, opened 2015-05-18
Change use of GCE secrets.py

dcarley:
#### Use GCE account JSON instead of PEM

Support for reading the JSON (as downloaded from Google) directly instead of
the extracted PEM was added in libcloud 0.17.0 (which is listed in our
`requirements.txt`:
- apache/libcloud@ab7d07b
#### Move `secrets.py` to root of repo

This `gce.ini` which specified `libcloud_secrets = ~/.secrets.py` wasn't
having any effect because the file has to be called `secrets.py` due to the
`import` statement in the inventory script.

It just happened to be working for me because I had once placed it in my
current working directory and it compiled to `secrets.pyc` which superseded
any other file. The same may have been true of Jim.

In addition to renaming it, I'm also moving it to the root of the repo and
adding it to `.gitignore`, to make it specific to this project (in case
anyone uses `libcloud` for anything else) and to make it more obvious what
the interactions are. For this reason we no longer need `gce.ini`.

---

/cc @annashipman


annashipman:
:+1: Thanks

PR #44, opened 2015-05-20
Remove Vagrant

dcarley:
This is very speculative. I'd like some opinions from anyone that's using it before it gets merged.

---

I initially added this when we had a single shared environment could work on
stories locally without affecting other people. Now that we have the ability
to create our own environments for each story, we're using dynamic
inventories and making more use of features that only exist in cloud
providers (registry object storage, load balancer appliances, etc), this
isn't so useful anymore and we haven't been maintaining it. So remove it :(


annashipman:
:+1: for removing it if no-one is using it. I haven't been.

actionjack:
I used to use the vagrant environment a lot but I have no strong opinions on keeping it around.

annashipman:
No one has yelled about this, so merging.

PR #42, opened 2015-05-18
[Fixes #94658242] Wrap commands in Makefile

dcarley:
Bit off piste. This is related to dynamic inventory, but also fixes another problem that we had a story for.
#### Move ansible-playbook commands to Makefile

These are getting quite complex, so we should move them to something more
automated to reduce the chance of human error.

By doing so we can also check that the environment name is set.

I realise that this disadvantages people using Linux instead of Mac because
they are forced to use `vault_password.sh` instead of their own solutions
like reading from a file. It seems like a reasonable compromise though.
#### Render ssh.config based on environment name

`ssh_config(5)` doesn't support interpolation of environment variables, so
we have been updating this by hand for each independent environment. This is
a pain and will become worse when we need to do the same in Jenkins. So
render it based on a crude template and `gitignore` the resulting file.

PR #45, opened 2015-05-20
Allow additional args to be passed to Makefile

dcarley:
So that you can, for example, see verbose output from Ansible:

```
make aws DEPLOY_ENV=dcarley ARGS=-vvvv
```

/cc @annashipman this might help you debug the problem you had in #42


dcarley:
Just rebased this with a fix for a typo.

actionjack:
:+1:

PR #47, opened 2015-05-20
[#94939334] Librarian-ansible to ansible-galaxy refactor

actionjack:
[Manage all Ansible dependencies from Ansible Galaxy](https://www.pivotaltracker.com/story/show/94939334)

**What**

`Librarian-Ansible` is written in `Ruby` and building out the environment with both `Ruby` and `Python` dependencies is starting to cause pain. `Librarian-ansible` is also not well supported. `Ansible-galaxy` is part of the Ansible code and doesn't require any additional language libraries.

**How**

This pull request should be reviewed with the following context:
1. I want to create a `requirements.yml` file that will mirror the dependencies described in the `Ansiblefile`,
2. I want to show people how to use `ansible-galaxy` and show we no-longer use ruby or bundler,
3. I want to make it easy to use `ansible-galaxy` by putting it in a Makefile so I no longer forget or make mistakes on the command-line,
4. I want to remove all locally stored playbooks because I have moved them to separate repositories,
5. I want to remove all the support files and directories that `librarian-ansible` was using.

**Who should review this PR**
- Any one on the core team
- I paired with @dhilton and @jimconner so they can't review this PR :stuck_out_tongue:


requirements.yml
```diff
@@ -0,0 +1,30 @@
...
+- name: tsuru_api
+  src: https://github.com/alphagov/ansible-playbook-tsuru_api.git
+  version: v0.0.1
+- name: gandalf
+  src: https://github.com/alphagov/ansible-playbook-gandalf.git
+  version: v0.0.1
+
+# GDS Forked, pull-request sent upstream awaiting merge.
+- name: bennojoy.redis
+  src: https://github.com/alphagov/redis.git
```
dcarley: Could we take this opportunity to rename of fork of the repo to `ansible-playbook-redis`? I can see it confusing another team if they need to fork the main Redis at a later date.

README.md
```diff
@@ -53,9 +49,9 @@ ssl_crt: |
   -----END CERTIFICATE-----
 ```
 
-* Install or update ansible playbooks using [librarian-ansible](https://github.com/bcoe/librarian-ansible):
+* Install or update ansible playbooks using [ansible-galaxy](http://docs.ansible.com/galaxy.html):
 ```
-bundle exec librarian-ansible install
+ansible-galaxy install -r requirements.yml
```
dcarley: Do we need to list this at all now that it's handled by the `Makefile`?

dcarley:
We should add the two new targets to `.PHONY` at the top. Otherwise, if we ever happen to have a file called `ansible-galaxy`, then the target won't be run.

Should we also delete `.librarian`, `Ansiblefile`, `Ansiblefile.lock`? Plus remove `tmp` and `roles` from `.gitignore`?

requirements.yml
actionjack: @dcarley @annashipman is this possible?

README.md
actionjack: @dcarley: Not everyone can use the Makefile (i.e. not everyone uses a Mac :smiling_imp:

dcarley: Because they're using `--vault-password-file=password.txt` or something else?

I'm inclined to think that we should either fix that or refer people to the contents of the `Makefile`, because if we duplicate content in the `README` then one of them is likely to fall out-of-date.

actionjack:
@dcarley I think we still need roles in the .gitignore but I'll fix the others

dcarley:
> @dcarley I think we still need roles in the .gitignore but I'll fix the others

You're right :ok_hand:

README.md
annashipman: I agree that we should fix the Makefile so that everyone can use it. For now, refer to the contents, and I'll talk to Brett about raising a story.

requirements.yml
actionjack: Fixed.

annashipman: Fixed how? By changing the name of the repo? In that case we need to change the Gist as well otherwise any PRs won't show up on the dashboard.

I can do that, but I just want to flag up to everyone that this needs to happen when repos are renamed.

dcarley: Sorry, this happened in Slack. Very good point. We might not need to add this one to the dashboard though, because we shouldn't be raising PRs against our master branch and it'll go away when the upstream PR is merged?

dcarley:
Thanks @actionjack :+1: :+1:

requirements.yml
annashipman: You're right that we shouldn't be raising PRs against master, but I've added it to the dashboard anyway just in case we do for some reason; it's a potentially irritating bug later.

PR #48, opened 2015-05-20
Use LB for internal access to API

dcarley:
A regression was introduced in the dynamic inventory work (not a surprise
because it was very complex - alphagov/tsuru-ansible#41) which meant that we
stopped using the load balancer for internal access to the API and only used
the first instance.

For example on the Gandalf host:

```
ubuntu@ip-10-128-10-180:~$ grep TSURU_HOST /home/git/.bash_profile
export TSURU_HOST=http://10.128.13.209:8080
```

Re-purpose the values with new key names to use as a variable for delegating
to the first API node and re-purpose the key name to be the DNS name of the
load balancer. This has a different hostname on AWS which is dependent on
the renaming in alphagov/tsuru-terraform#52

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -60,7 +60,7 @@
     - name: generate tsuru token
       run_once: true
       shell: tsr token
-      delegate_to: "{{ tsuru_api_internal_lb }}"
+      delegate_to: "{{ tsuru_api_host }}"
       register: tsr_token
 
     - name: write tsuru token to .bash_profile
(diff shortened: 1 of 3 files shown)


dcarley:
Tested on AWS.

dcarley:
One sec, going to refactor.

dcarley:
No, I'm not..

mtekel:
Tested on AWS & works fine, merging.

PR #49, opened 2015-05-20
[#93690844] Address registry using DNS again

dcarley:
A regression was introduced in the dynamic inventory work (not a surprise
because it was very complex - alphagov/tsuru-ansible#41) which meant that we
went back to address the Docker registry by IP instead of hostname. This
causes problems when the IP of the registry machine is changed because
images are namespaced by the registry that they come from.

Go back to using DNS like we did in 42e3bfb1dc0bf1a7935917ff0d4e574ab29fe332
for the (then) static inventory.

I'm also removing the duplicate entry for `docker_registry_port`. We only
need the entry in globals, for both platforms.

### DIFF
--- a/group_vars/all/globals.yml
+++ b/group_vars/all/globals.yml
@@ -4,6 +4,7 @@ redis_port: 6379
 
 api_port: 8080
 
+docker_registry_host: "{{ deploy_env }}-docker-registry.{{ domain_name }}"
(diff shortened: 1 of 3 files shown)


dcarley:
Didn't need rebasing :beers:

jimconner:
LGTM - Merging.

PR #53, opened 2015-05-26
Update hostname to tag/regex format for dual-cloud deploy

jimconner:
It seems that we overlooked the postgres server name during our
rework for dual-cloud compatibility.


dcarley:
Raises the issue of how we detect machines that haven't been configured due to typos. Will talk to Brett about adding a story.

PR #55, opened 2015-05-26
Abstract difference in AWS/GCE tags

dcarley:
By using an intermediate variable. This doesn't allow us to de-dupe any
variables, but it does make the `hosts` entries much easier to read and mean
that we don't need to use regexps (which are error prone).

I've tested this with the following commands and verified that the lists of
hosts under each "play" remain the same:

```
time make aws DEPLOY_ENV=dcarley ARGS=--list-hosts
time make gce DEPLOY_ENV=dcarley ARGS=--list-hosts
```

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -1,33 +1,33 @@
 # Bring up db tier (redis, mongodb).
-- hosts: ~({{ deploy_env }}|tag_Name_{{ deploy_env }})-tsuru-db
+- hosts: "{{ hosts_prefix }}-tsuru-db"
   sudo: yes
   roles:
     - bennojoy.redis
 
-- hosts: ~({{ deploy_env }}|tag_Name_{{ deploy_env }})-tsuru-db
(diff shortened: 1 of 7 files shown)


mtekel:
I have tested this in AWS and it works correctly, GCE testing is blocked on https://github.com/alphagov/tsuru-ansible/pull/57 (I can only test one branch at a time).

dcarley:
Unblocked.

NB: I'd recommend testing using your own independent environment rather than Jenkins because as soon as Jenkins goes into service it should only ever deploy from the `master` branch, so that we have confidence about how the shared environment has been built.

mtekel:
I am using my own jobs to build my own environments. Could you re-base this on the master now that #58 is merged? Otherwise I still can't test...

dcarley:
Okay, rebased.

mtekel:
Works & I was able to deploy dashboard in GCE, merging.

PR #56, opened 2015-05-26
Replace curl commands with Ansible uri module

dcarley:
#### Install python-httplib2 on API nodes

I'm planning to use Ansible's uri module which depends on this. If we make
more extensive use of this in the future then we should consider installing
it on all nodes.
#### Replace curl commands with Ansible uri module

These shell-outs and curl commands weren't registering an error when the
admin user wasn't created and/or it was unable to login. It would write an
empty `~/.tsuru_token` file and a later task which adds a docker node would
hang indefinitely because it triggered an interactive login.

Fix this by using Ansible's uri module which checks status codes and allows
us to parse the JSON response without piping it through another command.
#### Don't error if admin user already exists

The API will return `409 Conflict` if the user already exists. Allow this to
happen and for the subsequent tasks to continue.

---

Tested on AWS, new and existing.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -48,14 +48,24 @@
 # TODO : handle the case where the user already exists
     - name: add admin user
       run_once: true
-      shell: >
-        curl -sS -XPOST -d"{\"email\":\"{{admin_user}}\",\"password\":\"{{admin_password}}\"}" "http://127.0.0.1:{{api_port}}/users"
+      uri:
+        method=POST body_format=json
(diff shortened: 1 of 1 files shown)


mtekel:
A rebase on #58 will make GCE testing much easier. AWS looks OK so far...

dcarley:
Rebased.

mtekel:
both aws and gce OK...

PR #57, opened 2015-05-26
[#93438736] Add support for ssh over bastion hosts for GCE

actionjack:
Ansible's `gce.py` dynamic inventory script defaults to using public
IP addresses but we use bastion hosts to connect and therefore we
need to use private IP addresses to prevent the following error:

```
fatal: [ci-tsuru-gandalf] => SSH Error: ssh: connect to host 104.155.72.198 port 22: Connection timed out
    while connecting to 104.155.72.198:22
It is sometimes useful to re-run the command using -vvvv, which prints SSH debug output to help diagnose the issue.
```

This enhances security as we only need ssh opened on a single host.

**What**

We have added the option to use the `ANSIBLE_USE_PRIVATE_SSH` environment variable (if it exists) to instruct gce.py to output the hosts `private IP` as the `ansible_ssh_host` value rather than the hosts `public IP`

**Who should review this PR**
- Anyone on the core team,
- I paired with @jimconner on this one so he has to keep his mitts off the merge button


gce.py
```diff
@@ -215,8 +215,12 @@ def node_to_dict(self, inst):
                 md[entry['key']] = entry['value']
 
         net = inst.extra['networkInterfaces'][0]['network'].split('/')[-1]
+        # If there is no public IP listed, set the public_ip field to a null.
         if inst.public_ips.__len__() == 0:
           inst.public_ips=[None]
+
+        # Use private IPs if we're told to, or we've got no choice.
+        if os.environ.has_key("ANSIBLE_USE_PRIVATE_SSH") or inst.public_ips==[None]:
```
dcarley, line 223: I'm not sure about the environment variable. Do we ever want to turn it off?

If not, we shouldn't support doing so because it will remain untested.

If we do, then I think we should add another key to the config file that it already uses, because it's easier to understand what it's being used by.

jimconner, line 223: The default from ansible is to use public_ips... If we break their default, there is little chance that they'll ever accept this pull upstream. The use of the environment variable was to ensure the default for ansible remained unchanged.

dcarley, line 223: I wouldn't expect them to accept the environment variable upstream when the script already uses a config file.

So I think we should either do the quick and dirty thing to change the default in our forked version, or do the right thing and extend the config to something that could be PR'ed upstream.

mtekel:
Tested with both defined and undefined variable and it works as expected. I also think the default behaviour should be unchanged (no variable, no change in behaviour = use external IP if available). Ok to merge?

PR #59, opened 2015-05-27
[#92502208] Add support for PGP-encrypted vault password.

jimconner:
Distributing the `ansible` vault password throughout the team is a needless challenge which we can avoid by using multi-key encryption via `GPG`. The `ansible-vault` is still encrypted using a single pass-phrase as before, but access to that pass-phrase can now be controlled through GPG.

This PR implements changes as follows:
- Addition of an `open_the_vault.sh` script which uses GPG to decrypt the vault pass-phrase and outpus it on stdout
- A change to the ansible.cfg to tell ansible to use the `open_the_vault.sh` script to get the vault decryption password
- Addition of a gpg.recipients file which lists KEY-ID and email address for each approved recipient
- Addition of a `import-gpg-keys` stanza to the `Makefile` which will import keys listed in `gpg.recipients` into the GPG keyring
- Addition of a `recrypt` stanza to the `Makefile` which will decrypt the vault, generate a new pass-phrase, PGP-encrypt the pass-phrase and then re-encrypt the vault
  
  Whenever the 'recrypt' function is called, a new vault pass-phrase is generated and GPG-encrypted so that it can be accessed by the users listed in `gpg.recipients`

**_Known Features**_
- Because the `import-gpg-keys` function is decoupled from the `recrypt` function, it is possible to attempt to re-encrypt the passphrase using a non-trusted key - This will fail and the vault will be left unencrypted as a result. Please be caresful to ensure that you've imported all the keys before re-encrypting.


Makefile
```diff
@@ -28,3 +28,10 @@ clean-roles:
 
 ansible-galaxy:
 	ansible-galaxy install -r requirements.yml --force
+
+import-gpg-keys:
+	$(foreach var,$(shell cat gpg.recipients | awk -F: '{print $$1}'),gpg --keyserver hkp://pgp.mit.edu --search-keys $(var);)
+
+recrypt:
+	ansible-vault decrypt group_vars/all/secure && pwgen -cynC1 15 | gpg --batch --yes --trust-model always -e -o vault_passphrase.gpg $(shell cat gpg.recipients | awk -F: {'printf "-r "$$1" "'}) && ansible-vault encrypt group_vars/all/secure
```
dcarley, line 36: What do you think about adding the following before this line to prevent the vault being left unencrypted if you don't have all of the keys?

``` make
  @$(foreach var,$(shell cat gpg.recipients | awk -F: '{print $$1}'), \
    gpg --list-public-keys "$(var)" >/dev/null 2>&1 || \
      (echo "Public key '$(var)' not found in your keyring"; exit 1); \
  )
```

dcarley:
Could you include some words in 919ee8d4b17ac6fa555b886c3ef244e0b6b43317 about why we want to do this? Ability to add new people without distributing password and ability to remove people so that they don't see future secrets (but no mitigation for existing secrets).

README.md
```diff
@@ -22,6 +22,96 @@ GCE_PARAMS = ('...@developer.gserviceaccount.com', '/path/to/gce_account.json')
...
+
+```
+sudo apt-get update
+sudo apt-get install pwgen
+sudo apt-get install gnupg2
+sudo apt-get install gnupg-agent
+sudo apt-get install pinentry-curses
+```
+
+##### Common
```
dcarley, line 53: Is there any third-party documentation that we can link to instead of having it here? Makes our README very long and we might not always have the most up-to-date information. Not a deal breaker though.

Makefile
```diff
@@ -28,3 +28,10 @@ clean-roles:
 
 ansible-galaxy:
 	ansible-galaxy install -r requirements.yml --force
+
+import-gpg-keys:
+	$(foreach var,$(shell cat gpg.recipients | awk -F: '{print $$1}'),gpg --keyserver hkp://pgp.mit.edu --search-keys $(var);)
```
dcarley: We could use macros for `$(shell cat gpg.recipients | awk -F: '{print $$1}')` and `group_vars/all/secure` (below) since we use them more than once and the length makes this quite difficult to read.

dcarley:
Nice work :+1: :+1:

dcarley:
Oh, do we also need to generate a keypair for `deploy.tools` and add it to the recipients now?

jimconner:
Thats a WIP currently @dcarley - We've made the keys for jenkins and are just working on modifications to multicloud-deploy so that the Jenkins box will get it's keys when provisioned.

Makefile
jimconner: It is only group_vars/all/secure which is used twice - the two awk lines are different.

Makefile
jimconner, line 36: The Makefile has now been modified so that the import step is always run before the recrypt one.

jimconner:
GPG key for jenkins now added to the recipients, and the vault has been re-encrypted for use by Jenkins user.

dcarley:
The import target seems to contact the keyserver and prompt me for keys that I already have, when I don't think it should do either:

```
(ansible)➜  tsuru-ansible git:(pgp_keyserver_support) ✗ gpg --list-keys BCCB8A5F
pub   2048R/BCCB8A5F 2013-06-20
uid                  Anna Shipman <anna.shipman@digital.cabinet-office.gov.uk>
sub   2048R/E984D402 2013-06-20
(ansible)➜  tsuru-ansible git:(pgp_keyserver_support) ✗ make recrypt
gpg --keyserver hkp://pgp.mit.edu --search-keys BCCB8A5F; gpg --keyserver hkp://pgp.mit.edu --search-keys 22652456; gpg --keyserver hkp://pgp.mit.edu --search-keys 77AACBC2; gpg --keyserver hkp://pgp.mit.edu --search-keys B81D5E45; gpg --keyserver hkp://pgp.mit.edu --search-keys F35E9C9E;
gpg: searching for "BCCB8A5F" from hkp server pgp.mit.edu
(1)     Anna Shipman <anna.shipman@digital.cabinet-office.gov.uk>
          2048 bit RSA key BCCB8A5F, created: 2013-06-20
Keys 1-1 of 1 for "BCCB8A5F".  Enter number(s), N)ext, or Q)uit >
```

PR #60, opened 2015-06-01
Refactor Makefile

dcarley:
No functional changes.

---
#### Makefile: Hard wrap long lines …

So that they're easier to read. The groupings in `.PHONY` are logical.
#### Makefile: Move common deps to "preflight" target …

So that it's easy to see what dependencies are common between aws/gce.
#### Makefile: break AWS check to separate target …

To match `check-env-var` and make the gce/aws targets look similar.
#### Makefile: move ansible-playbook command to macro …

So that we don't duplicate the argument structure between aws/gce.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,13 @@
 .PHONY: all \
 	aws gce \
-	check-env-var render-ssh-config \
+	preflight check-env-var render-ssh-config \
 	clean-roles ansible-galaxy \
 	import-gpg-keys recrypt
 
 all:
 	$(error Usage: make <aws|gce> DEPLOY_ENV=name [ARGS=extra_args])
 
-aws: check-env-var render-ssh-config
+aws: preflight
 ifndef AWS_SECRET_ACCESS_KEY
 	$(error Environment variable AWS_SECRET_ACCESS_KEY must be set)
 endif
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -30,8 +44,13 @@ ansible-galaxy:
...
+		$(shell cat gpg.recipients | awk -F: '{print $$1}'), \
+		gpg --list-public-key $(var) || gpg --keyserver hkp://keyserver.ubuntu.com --search-keys $(var); \
+	)
 
 recrypt: import-gpg-keys
-	ansible-vault decrypt group_vars/all/secure && pwgen -cynC1 15 | gpg --batch --yes --trust-model always -e -o vault_passphrase.gpg $(shell cat gpg.recipients | awk -F: {'printf "-r "$$1" "'}) && ansible-vault encrypt group_vars/all/secure
-
+	ansible-vault decrypt group_vars/all/secure && pwgen -cynC1 15 | \
+		gpg --batch --yes --trust-model always -e -o vault_passphrase.gpg \
+			$(shell cat gpg.recipients | awk -F: {'printf "-r "$$1" "'}) && \
```
actionjack, line 55: @dcarley Did you want that extra tab there?

dcarley, line 55: Yeah, it was meant to indicate that it's a continuation of the command on the last line, rather than a new command in the pipe/chain. Is it more confusing than helpful?

actionjack:
LGTM Merging.

PR #61, opened 2015-06-01
[#techdebt] Add Michal Tekel's gpg key id and re-encrypting vault

actionjack:
Adding Michal's gpg key id to the gpg.recipients file and re-encrypting the vault.


mtekel:
This works, thanks.

PR #63, opened 2015-06-02
Remove unnecessary comments

annashipman:
This is a good example of why comments should be used sparingly, if
at all; the comment in `site-gce.yml` was incorrect and has been for
as long as this file has existed. However, neither comment is
required, as the file names alone are clear enough indicators of
what the files are for.


mtekel:
I agree. A good second look/peer review can spot things you missed before...

PR #66, opened 2015-06-03
[#94300924] Add post install deploy tasks

mtekel:
- specify pool name when registering docker node (fixes tsuru 0.11.1 behaviour where nodes are not added to any pool by default)
- configure default pools
- add insecure_deployer key for git
- install python platform
- deploy dashboard
- deploy postgresapi
- configure postgresapi
- display app info in the end of run

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -104,3 +104,4 @@
 
 - include: postgres.yml
 - include: ssl-proxy.yml
+- include: post-install.yml
--- a/common.yml
+++ b/common.yml
@@ -93,12 +93,12 @@
   post_tasks:
     - name: Register node.
       shell: >
-        tsuru-admin docker-node-add --register address=http://{{ gce_private_ip }}:{{ docker_port }}
+        tsuru-admin docker-node-add --register address=http://{{ gce_private_ip }}:{{ docker_port }} pool=default
       delegate_to: "{{ tsuru_api_host }}"
       when: gce_private_ip is defined
     - name: Register node.
       shell: >
-        tsuru-admin docker-node-add --register address=http://{{ ec2_private_ip_address }}:{{ docker_port }}
+        tsuru-admin docker-node-add --register address=http://{{ ec2_private_ip_address }}:{{ docker_port }} pool=default
       delegate_to: "{{ tsuru_api_host }}"
       when: ec2_private_ip_address is defined
 
--- a/platform-aws.yml
+++ b/platform-aws.yml
@@ -1,6 +1,8 @@
 ---
 ip_field_name: "ec2_private_ip_address"
 domain_name: "tsuru.paas.alphagov.co.uk"
+api_protocol: "https"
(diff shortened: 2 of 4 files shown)


annashipman:
I get

```
failed: [annashipman-tsuru-api-1] => {"elapsed": 60, "failed": true}
msg: Timeout when waiting for 127.0.0.1:8080
failed: [annashipman-tsuru-api-0] => {"elapsed": 60, "failed": true}
msg: Timeout when waiting for 127.0.0.1:8080

FATAL: all hosts have already failed -- aborting
```

mtekel:
Is this your first deploy? Or an update to existing environment? We've seen this issue on AWS, when we temporarily used API_PORT variable for st. else and thus forced port 443 on tsr api server, which wouldn't start... Which port do you have in the tsuru api config? (I think it's in /etc/tsuru/...)

dcarley:
Might be the problem fixed by 0cfaa77256f8e9f8354ffd4582e0317c29ec70b5 and ansible-playbook-tsuru_api@eea273c244eff4fa4d0da9a4a0b549769da0ae98.

Could you try running  `make clean-roles ansible-galaxy` to update the dependencies?

If that works, we might want to reconsider adding that to the default target.

annashipman:
@mtekel it was a new environment.

@dcarley that worked. We could add that to the default target, but that will make it take quite a bit longer. I'll raise a PR and we can discuss there.

Meanwhile, I've now successfully tested this and it's fine, so merging.

annashipman:
Gah! Because I had the error, I checked out master to see if it worked there, and then when re-testing I completely forgot and tested master... which worked!

I only realised when CI failed, so I have raised a PR to revert it: https://github.com/alphagov/tsuru-ansible/pull/67

The error is

```
TASK: [Tsuru target-list] *****************************************************
failed: [localhost] => {"changed": true, "cmd": "tsuru target-list", "delta": "0:00:00.001938", "end": "2015-06-04 15:37:33.904116", "rc": 127, "start": "2015-06-04 15:37:33.902178", "warnings": []}
stderr: /bin/sh: 1: tsuru: not found
```

mtekel:
Yes, jenkins needs to install tsuru agent (another reason I prefer testing on jenkins rather than locally). Also, I think ansible galaxy clean could/should be in the default target (unless ansible galaxy automatically updates dependencies when requirements are updated, which it seems not to). It adds a bit more time, but it's very short time compared to how long whole deployment takes...

annashipman:
This isn't to do with whether it was tested on Jenkins or locally; it also fails locally with the same error.

dcarley:
Would it be safer/faster to run these commands on the API server, which already has the tsuru commands and a login token?

mtekel:
Yes and that would also solve problem of two concurrently running jobs trying to be logged it to two different environments at the same time...

PR #68, opened 2015-06-05
Add re-fetching ansible galaxy roles to default

annashipman:
It is easy to forget to do this and then get unexplained errors. Adding this to the default target only takes a few extra seconds.

As discussed in https://github.com/alphagov/tsuru-ansible/pull/66

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ aws: preflight check-env-aws
 gce: preflight
 	SSL_CERT_FILE=$(shell python -m certifi) $(call ANSIBLE_PLAYBOOK_CMD,gce.py,gce)
 
-preflight: check-env-var render-ssh-config
+preflight: check-env-var render-ssh-config clean-roles ansible-galaxy
 
 check-env-aws:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,10 +13,10 @@ ANSIBLE_PLAYBOOK_CMD = ansible-playbook \
 all:
 	$(error Usage: make <aws|gce> DEPLOY_ENV=name [ARGS=extra_args])
 
-aws: preflight check-env-aws
+aws: preflight check-env-aws clean-roles ansible-galaxy
```
dcarley: We could add these to the `preflight` target to prevent the duplication in the `aws` and `gce` targets.

annashipman: Oh yes, good point. Will do.

annashipman: Rebased to do this thing which I should have done in the first place!

dcarley:
This might make it more difficult to develop changes to playbooks. I guess we'll see how it goes and adapt it as necessary, because I think it's more important to be consistent.

PR #71, opened 2015-06-08
[#95697144] post install tasks

mtekel:
- specify pool name when registering docker node (fixes tsuru 0.11.1 behaviour where nodes are not added to any pool by default)
- configure default pools
- add insecure_deployer key for git
- install python, java and static platforms
- deploy dashboard
- deploy postgresapi
- configure postgresapi
- display app info in the end of run

All tsuru operations are run on tsuru-api*[0] server so that GCE and AWS jobs can run concurrently. Code of dashboard and postgresapi is pushed from localhost.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -115,3 +115,4 @@
 
 - include: postgres.yml
 - include: router.yml
+- include: post-install.yml
--- a/post-install.yml
+++ b/post-install.yml
@@ -1,0 +1,158 @@
+- hosts: "{{ hosts_prefix }}-tsuru-api*[0]"
+  sudo: yes
+  tasks:
+    - name: Check tsuru_deployer key
+      stat: path=~/.ssh/tsuru_deployer
+      register: tsuru_deployer_key
+      delegate_to: "localhost"
+      sudo: no
+
+    - name: Generate tsuru_deployer key
+      shell: >
+        ssh-keygen -b 2048 -t rsa -f ~/.ssh/tsuru_deployer -q -N ""
+      when: tsuru_deployer_key.stat.exists == False
+      delegate_to: "localhost"
+      sudo: no
+    - name: Generate tsuru_deployer.pub key
+      shell: >
+        ssh-keygen -y -f ~/.ssh/tsuru_deployer >~/.ssh/tsuru_deployer.pub
+      when: tsuru_deployer_key.stat.exists == False
+      delegate_to: "localhost"
+      sudo: no
+
+    - name: Get localhost home
+      shell: >
+        echo ~
+      register: local_home
+      delegate_to: "localhost"
+      sudo: no
+    - name: Deploy tsuru_deployer.pub key
+      copy:
+        src: "{{ local_home.stdout }}/.ssh/tsuru_deployer.pub"
+        dest: ~/.ssh/
+        mode: 0600
+
+    - name: Tsuru pool-list
+      shell: > 
+        tsuru-admin pool-list
+      register: tsuru_pools
+    - name: Tsuru pool add default
+      shell: >
+        tsuru-admin pool-add default
+      when: "not 'default' in tsuru_pools.stdout"
+    - name: Check admin pool
+      shell: >
+        tsuru pool-list | grep admin
+      register: admin_pool
+    - name: Add admin team to default pool
+      shell: >
+        tsuru-admin pool-teams-add default admin
+      when: "not 'default' in admin_pool.stdout"
+
+    - name: Tsuru platform-list
+      shell: >
+        tsuru platform-list | awk {'print $2'}
+      register: tsuru_platforms
+    - name: Tsuru platform add python
+      ignore_errors: no
+      shell: >
+        tsuru-admin platform-add python -d https://raw.github.com/tsuru/basebuilder/master/python/Dockerfile
+      when: "not 'python' in tsuru_platforms.stdout"
+
+    - name: Tsuru key-list
+      shell: >
+        tsuru key-list
+      register: tsuru_keys
+    - name: Tsuru add ssh key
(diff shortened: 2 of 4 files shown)


mtekel:
One more note: This depends on https://github.com/alphagov/tsuru-terraform/pull/60

common.yml
```diff
@@ -88,14 +88,15 @@
   post_tasks:
     - name: Register node.
       shell: >
-        tsuru-admin docker-node-add --register address=http://{{ gce_private_ip }}:{{ docker_port }}
+        tsuru-admin docker-node-add --register address=http://{{ gce_private_ip }}:{{ docker_port }} pool=default
```
dcarley: Will this need to be applied manually to the CI environment with `docker-node-update`?

post-install.yml
```diff
@@ -0,0 +1,146 @@
...
+        tsuru env-set -a postgresapi POSTGRESAPI_SHARED_ADMIN_PASSWORD={{ pg_admin_pass }};
+        tsuru env-set -a postgresapi POSTGRESAPI_SHARED_PUBLIC_HOST={{ postgres_host }};
+      when: "not 'Everything up-to-date' in postgresapi_push_result.stderr"
+    - name: Tsuru app-run db upgrade
+      shell: >
+        tsuru app-run --app postgresapi -- python manage.py upgrade_db
+      when: "not 'Everything up-to-date' in postgresapi_push_result.stderr"
+
+    - debug: var=dashboard_app_info.stdout_lines
+    - debug: var=postgresapi_app_info.stdout_lines
```
dcarley, line 157: Are these intentionally left in?

jimconner, line 157: Yes - it's useful to have the dashboard and postgresapi URLs shown at the end of the run.

post-install.yml
```diff
@@ -0,0 +1,146 @@
...
+        tsuru target-list
+      register: tsuru_targets
+    - name: Tsuru add target.
+      shell: >
+        tsuru target-add {{ deploy_env }}_{{ domain_name }} http://127.0.0.1:8080
+      when: "not '{{ deploy_env }}_{{ domain_name }}' in tsuru_targets.stdout"
+    - name: Tsuru set target.
+      shell: >
+        tsuru target-set {{ deploy_env }}_{{ domain_name }}
+    - name: Tsuru login
```
dcarley: Are these steps necessary? Do we have another playbook that does this?

post-install.yml
```diff
@@ -0,0 +1,146 @@
...
+      shell: >
+        echo {{ admin_password }} | tsuru login {{ admin_user }}
+
+    - name: Deploy insecure-deployer.pub key
+      copy:
+        src: files/insecure-deployer.pub
```
dcarley: Not sure about proliferating the use of this key. Should we generate a new keypair within the environment?

post-install.yml
jimconner: The common.yml playbook is doing these login steps as the root user, whereas we're not sudo'd for these steps.

dcarley: Is there a reason the other one is sudo? Should they either both be sudo or not sudo?

jimconner: The best reason I could think of for the other one using sudo is the apt package installation - I'm not sure if it's actually required for apt installs - It's certainly not needed for the rest of the steps being carried out on that host.

post-install.yml
mtekel: This is the public key, which should be safe to share with the word. Requirement here is that this is the same key as we (or jenkins) have locally, so that we can do git push. Generating new key every time would mean we/jenkins would have to deploy a new key for every new environment...

common.yml
```diff
@@ -88,14 +88,17 @@
   post_tasks:
     - name: Register node.
       shell: >
-        tsuru-admin docker-node-add --register address=http://{{ gce_private_ip }}:{{ docker_port }}
+        tsuru-admin docker-node-add --register address=http://{{ gce_private_ip }}:{{ docker_port }} ;
+        tsuru-admin docker-node-update http://{{ gce_private_ip }}:{{ docker_port }} pool=default
```
dcarley: Could you break these out to separate tasks?

We've had problems with multiple shell commands in single tasks. Ansible only reports the exit code of the last command, so it makes it quite hard to tell if something has gone wrong and what it was.

post-install.yml
dcarley: > Generating new key every time would mean we/jenkins would have to deploy a new key for every new environment...

Is that a bad thing? We could use our personal keys and Jenkins could use its own (which would be generated on and never leave the Jenkins server).

I'd like to stop using the `insecure-deployer` keypair because we've shared it amongst many users and have no way of individually revoking them.

post-install.yml
dcarley: Can we combine the two then?

mtekel:
Updated key handling, unified tsuru account handling with common.yml (use same user, targets etc.)

post-install.yml
```diff
@@ -0,0 +1,168 @@
...
+      delegate_to: "localhost"
+      sudo: no
+
+    - name: Generate tsuru_deployer key
+      shell: >
+        ssh-keygen -b 2048 -t rsa -f ~/.ssh/tsuru_deployer -q -N ""
+      when: tsuru_deployer_key.stat.exists == False
+      delegate_to: "localhost"
+      sudo: no
+    - name: Generate tsuru_deployer.pub key
```
actionjack, line 14: This feels fairly intrusive, can we use some other way? Generating keys on my laptop for remote applications with out my approval feels like we are walking in root kit territory.

mtekel:
Now pushing from API server.

actionjack:
LGTM

PR #73, opened 2015-06-09
Add gpg keys for new team members

actionjack:
Add gpg public keys for:
- Colin Saliceti
- Dan Hilton
- Hector Rivas


jimconner:
LGTM - Merging

PR #74, opened 2015-06-09
[#93541780] Nginx SSL reverse proxy in front of API

dcarley:
So that we can secure access to the API on all providers.

We don't currently have SSL on the API in GCE environments because the load
balancers which do SSL termination aren't currently supported by Terraform.
We do have ELBs terminating SSL in AWS, but the certificate upload process
is outside of Terraform.

This won't currently be used. I'm going to roll this out in several steps
because we don't currently have an SSL certificate for GCE and the Tsuru
client won't work on cert validation errors. These steps will follow:
- change the AWS load balancer to forward to TCP:443 instead of terminating
  SSL (but retain HTTP:8080 forwarding for the moment)
- add TCP:443 forwarding to the GCE load balancer

Then when we have the GCE certificate deployed to the right machines:
- remove all HTTP:8080 listeners and forwarders on load balancers
- change the API to only listen on `127.0.0.1:8080`

I was originally hoping to implement this using Tsuru's `use-tls` option
which allows you to pass a certificate and key for Tsuru to do SSL
termination. However this would have made the local `tsuru` and
`tsuru-admin` commands fail, which hit the API over loopback to do things
like register Docker nodes, due to SSL certificate hostname validation.

That could have been overcome by calling out to the internal load balancer
but that would have required waiting for enough health checks to pass for
the instances to be brought into service and it all seemed like unnecessary
complexity.

### DIFF
--- a/ssl_proxy_vars.yml
+++ b/ssl_proxy_vars.yml
@@ -4,7 +4,7 @@ nginx_sites:
     - listen 443 ssl
     - ssl_certificate wildcard.tsuru.paas.alphagov.co.uk.site.chain.crt
     - ssl_certificate_key wildcard.tsuru.paas.alphagov.co.uk.key
-    - server_name tsuru.paas.alphagov.co.uk
+    - server_name {{ domain_name }}
     - location / {
       proxy_pass http://localhost:{{ upstream_port }}/;
       proxy_redirect default;
--- a/ssl_proxy_vars.yml
+++ b/ssl_proxy_vars.yml
@@ -1,0 +1,24 @@
+---
+nginx_sites:
+  ssl_reverse_proxy:
+    - listen 443 ssl
+    - ssl_certificate wildcard.tsuru.paas.alphagov.co.uk.site.chain.crt
+    - ssl_certificate_key wildcard.tsuru.paas.alphagov.co.uk.key
+    - server_name tsuru.paas.alphagov.co.uk
+    - location / {
+      proxy_pass http://localhost:{{ upstream_port }}/;
+      proxy_redirect default;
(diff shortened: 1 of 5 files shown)


dcarley:
Do you want the commit message rebased to reflect that we now have a certificate for GCE per #76?

jimconner:
Sorry, but yes - I guess the commit message should be tweaked due to the order these PRs are getting merged in.

ssl_proxy_vars.yml
```diff
@@ -0,0 +1,24 @@
+---
+nginx_sites:
+  ssl_reverse_proxy:
+    - listen 443 ssl
+    - ssl_certificate wildcard.tsuru.paas.alphagov.co.uk.site.chain.crt
+    - ssl_certificate_key wildcard.tsuru.paas.alphagov.co.uk.key
+    - server_name tsuru.paas.alphagov.co.uk
```
jimconner: We're going to need to do something about setting the server_name correctly to match the certificate name - I should have noticed that in my own SSL cert PR just now.

dcarley: The other story has been marked as finished. @keymon @actionjack are you taking care of it in this PR?

actionjack: Yes we're going to take care of it, I was initially wondering if you had some reasoning behind the current name but we'll take it from here if there wasn't.

dcarley: It was moved as-is from the router (previously sslproxy). I suspect the `server_name` probably doesn't matter, because it's the only vhost listening on 443 so it will receive all requests regardless of `Host` header. We should make it factually correct though.

PR #76, opened 2015-06-10
[#93542594] Use the correct SSL cert for different providers

jimconner:
**What**

We now have seperate SSL certificates for different cloud providers, so we no longer want to deploy the same certificate on both platforms.
- The vault has been updated with the additional certificate (gce_ssl_key/cert) and the original ssl_key/cert files have been renamed to aws_ssl_key/cert accordingly.
- We are defining which of the two certificates should be used inside platform-aws.yml and platform-gce.yml respectively

**How to review this PR**

I've created the PR to be reviewed in the following context:
- I need to install ensure that the SSL certificate which is deployed matches the domain name used for our deployments on each different cloud provider platform

**Who should review this PR**
- Anyone in the core team.


dcarley:
This will require a rebase of #74 

Is it possible to reduce the duplication by using an intermediate variable for each platform? e.g.

```
platform-aws.yml:
ssl_crt: "{{ aws_ssl_crt }}"

platform-gce.yml:
ssl_crt: "{{ gce_ssl_crt }}"
```

jimconner:
I don't think we can do that because the we can double-de-reference variables like this - We have the same thing with ip_field_name in the past and had to define lots of workarounds for the individual host names in the platform-xxx.yml files. I'm not certain if it was the double-de-referencing, or the fact that the different yml files are parsed at different stages (so gce_ssl_crt is not available when the platform-xxx.yml is parsed) - I'm pretty sure we can't do that though.

dcarley:
I think this might be different to `ip_field_name` which was needed to evaluate another variable name and more like `hosts_prefix` in aa2a244d0e5db3a04b80f66a2f57bf7489665456. Could you give it a quick try?

jimconner:
Seems to work OK as you suggested, so I've pushed a new commit over the top of the old one.

router.yml
```diff
@@ -8,8 +8,8 @@
      file: dest=/etc/nginx mode=755 owner=root group=root state=directory
    - name: Upload Unencrypted Key File
      copy: content="{{ ssl_key }}" dest=/etc/nginx/wildcard.tsuru.paas.alphagov.co.uk.key owner=root group=root
-   - name: Upload Unencrypted Cert File
-     copy: content="{{ ssl_crt }}" dest=/etc/nginx/wildcard.tsuru.paas.alphagov.co.uk.site.chain.crt owner=root group=root
+   - name: Upload Unencrypted Key File
+     copy: content="{{ ssl_key }}" dest=/etc/nginx/wildcard.tsuru.paas.alphagov.co.uk.key owner=root group=root
```
dcarley: Two keys? :scream:

dcarley:
> Seems to work OK as you suggested, so I've pushed a new commit over the top of the old one.

That's made me very happy :+1:

router.yml
jimconner: Now fixed - new commit pushed over the old one.

dcarley:
:tada:

PR #77, opened 2015-06-10
[#93705168] Configure docker registry to use gcs bucket

actionjack:
[Registry uses GCS on GCE](https://www.pivotaltracker.com/story/show/93705168)

**What**

**Acceptance Criteria**
**Given:** that I am deploying the platform to GCE
**When:** I run the provisioning and orchestration scripts
**Then:** the resulting platform has the Registry data stored on resilient storage outside the container
**When:** the Registry instance fails and a new Registry instance is created
**Then:** the new Registry instance is able to read the data on the resilient data store.

I've created this PR to be reviewed in the following context:
- I want to configure the registry to use a `gcs` bucket with interoperability keys,
- I want to show people how to configure their own environments,
- I have been made aware that the use of interoperability keys requires additional work for people to do which may not be appreciated, i.e. generating and managing your own keys,
- I need to pull in an updated playbook from upstream that will use service account based authorisation to share `gcs` bucket credentials across the entire team,
- I need to migrate my docker registry's `gcs` bucket authentication and authorisation method from interoperability keys to service account based authorisation,
- I need to update the documentation to show we have changed from interoperability keys to service account based authorisation.

**Who should review this PR**
- Any one on the main team with the exception of @dhilton, @keymon, Colin Saliceti since they paired with me on this one. :p

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,7 @@ root directory of this repo. The contents are as follows:
 ```
 GCE_PARAMS = ('...@developer.gserviceaccount.com', '/path/to/gce_account.json')
 GCE_KEYWORD_PARAMS = {'project': 'project_id'}
-GCS_ACCESS_KEY=<your gcs access key>
-GCS_SECRET_KEY=<your gcs secret key>
 ```
-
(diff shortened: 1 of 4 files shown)


jimconner:
I can find no fault whatsoever with this. I have tested this against GCE and confirmed it is working. Great PR. Merging.

dcarley:
:+1: :+1: :+1:

PR #79, opened 2015-06-10
Run ansible-galaxy only when necessary

keymon:
Currently we run the ansible-galaxy every time we run make,
even when it is not required if `requirements.yml` did not
change.

In this commit using a file as a flag to keep track of the
last run of `ansible-galaxy` task. `ansible-galaxy` will only
run if the task never run or `requirements.yml` is newer.

@annashipman originally added the this task.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 secrets.py*
 roles
 ssh.config
+.ansible-galaxy.check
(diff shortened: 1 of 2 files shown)


dcarley:
yey, thanks!!

annashipman:
Late to the party here, but :+1:

PR #80, opened 2015-06-10
Merge clean-roles and ansible-galaxy targets

dcarley:
So that we don't delete the roles on every run now that we only run
`ansible-galaxy` when the `requirements.yml` changes. Somehow this worked
fine when I tested #79 and then stopped working shortly after.


keymon:
+1 thx. Sorry I miss that one.

PR #81, opened 2015-06-10
De-dupe docker-node-* tasks for AWS and GCE

dcarley:
By getting the hosts private IP address from Ansible's default variables
(facts?) instead of the dynamic inventory. This has the same effect, but is
much simpler and means that we don't have to maintain the same commands in
two places.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -88,30 +88,18 @@
   post_tasks:
     - name: Register node.
       shell: >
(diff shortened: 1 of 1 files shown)


mtekel:
LGTM, works. We have also attempted to do st. similar, using `ansible_ssh_host`, but ansible wouldn't recognize that, so we created the branches...

PR #82, opened 2015-06-11
Use ssh.config when connecting to the nat host

keymon:
We use a custom `ssh.config` file to connect to our VM instances via
a bastion ssh host (nat). Ansible refers to this ssh.config file for
all instances.

But the private instances use the `ProxyCommand` and spawn a separated
ssh command, but in this case we don't pass the option `-F ssh.config`.

For consistency and to ensure that the config can run in any system,
we should add this option.


jimconner:
Works great for me - stops ssh from prompting me multiple times during deployment - Merging..

PR #83, opened 2015-06-12
[#95695338] Fix intermediate SSL certificate for wildcard

actionjack:
[Shut down insecure Tsuru API](https://www.pivotaltracker.com/story/show/95695338)

**What**

I need to replace the intermediate certificate for `*.tsuru2.paas.alphagov.co.uk` on gce. 
The current intermediate certificate give the following error when `gandalf` attempts to contact the `tsuru api` server :

```
$ curl https://tsuru-api.tsuru2.paas.alphagov.co.uk
curl: (60) SSL certificate problem: unable to get local issue
certificate
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
```


jimconner:
Having checked in the vault, I can see that the intermediate cert for tsuru2 now matches the one for AWS - I'm happy with this as I'd already tested that out yesterday... Merging.

PR #85, opened 2015-06-12
[#96756584 ] Configure graphite

mtekel:
Install and configure graphite server. Configure postgresapi and dashboard apps to use it.

[Tsuru metrics](http://docs.tsuru.io/en/latest/advanced_topics/metrics.html) use StatsD to transport data to StatsD server, which then in turn feeds them to graphite. Graphite is accessed by dashboard app to retrieve the metrics and show them to the user. This PR:
- installs and configures StatsD on graphite server
- installs and configures Graphite (including data retention and cleanup) on graphite server
- sets up environment variables of `postgresapi` and `dashboard` apps
- updates README about how to configure graphite credentials


group_vars/all/globals.yml
```diff
@@ -15,3 +15,7 @@ gandalf_port: 8080
 
 mongodb_conf_bind_ip: 0.0.0.0
 mongodb_port: 27017
+
+graphite_secret_key: "changeme"
+graphite_auth_user: "foo"
+graphite_auth_pass: "bar"
```
jimconner, line 21: Shouldn't these user credentials be inside the vault?... with the readme.md updated to reflect that they're required contents for the vault.

post-install.yml
```diff
@@ -102,6 +102,16 @@
...
+      register: dashboard_push_result
+
+    - name: Tsuru set environment variables for dashboard
+      shell: >
+        tsuru env-set -a dashboard STATSD_PORT=8125;
+        tsuru env-set -a dashboard STATSD_HOST={{ deploy_env }}-metrics.{{ domain_name }};
+        tsuru env-set -a dashboard GRAPHITE_HOST={{ deploy_env }}-metrics.{{ domain_name }};
+        tsuru env-set -a dashboard GRAPHITE_USER={{ graphite_auth_user }}
+        tsuru env-set -a dashboard GRAPHITE_PASS={{ graphite_auth_pass }}
+      when: "not 'Everything up-to-date' in dashboard_push_result.stderr"
```
jimconner, line 114: If we set these variables before we first push the dashboard, then it won't need to restart the dashboard after it's been deployed for it to read the newly defined variables - can we move them up to happen before the git push ?

common.yml
```diff
@@ -10,6 +10,8 @@
   roles:
     - greendayonfire.mongodb
 
+- include: tsuru_graphite.yml
+
```
jimconner, line 14: Is there a reason why this include appears between mongo and gandalf? Is it required by subsequent steps, or could it possibly be moved down to the end after all the standard tsuru components have been deployed (i.e. after router, but before post-install) ? - It would seem more logical in my mind that way as it's an add-on for use by tsuru rather than a core component.

mtekel, line 14: I just put it there because there was a comment about bringing up DB tier. Metrics and graphite are kind-of data. But it can be moved anywhere, it's just a line...

post-install.yml
mtekel, line 114: Yes, this in general needs proper fix. The variables need to be set every time they are changed. This is unrelated to the app deploys, which is what is now checked. Proposed way: output them to file and on a file change, actually run that file which will in turn set the variables...

jimconner, line 114: My point was that we can avoid a un-necessary restart of the dashboard app by having the variables defined before the dashboard is started for the first time.

group_vars/all/globals.yml
mtekel, line 21: I just this so that I could leave it to @dhilton to define the final values in the vault... It can be definitely removed and requested to be defined in vault (or on command line). This is more of a convenience feature, where things work immediately out of the box...

tsuru_graphite.yml
```diff
@@ -0,0 +1,47 @@
...
+    graphite_storage_file_retention_days: "65"
+  pre_tasks:
+    - name: Update apt cache
+      apt: update_cache=yes cache_valid_time=86400
+  roles:
+    - nsg.graphite
+    - dmichel1.statsd
+  tasks:
+    - name: Configure statsD idlestats
+      lineinfile:
```
keymon, line 15: I would suggest use templates for this configuration.

Inline edition is often error prone and potentially not idempotent as it depends in the current state!

mtekel, line 15: The catch is that these templates are defined in nginx and statsD roles. And what I need to change is currently not configurable. I would have to update both roles, do PR and wait for the authors to merge, which can take quite some time...

keymon:
In the PR  the `STATSD_HOST` and `STATSD_PORT` are explicitly set per application. If the platform user does not specify this variables, he won't get any metrics.

But is this the intended behaviour in our platform? Or do we want to get the metrics without having to set these two environment variables.

If it is the second case, we have to do a different approach. 

Possible solutions could be:
- Define the environment variables globally in tsuru. There is a feature request https://github.com/tsuru/tsuru/issues/932 related to this.
- Setup a statsd relay in each one of the docker hosts, so it listens in `localhost` which is the [default destination for the metrics](https://github.com/tsuru/tsuru/blob/8c2bb1ee3845e21c81f561d5292a495e430792ab/docs/advanced_topics/metrics.rst#sending-metrics), which would forward to the default graphite server.

tsuru_graphite.yml
keymon, line 15: In that case the file would be changing back and forward every time we run ansible, restarting the services each time.

I know that is more work, but I would say that we fork and do a PR, and we point to our own fork for the time being.

annashipman, line 15: I'm inclined to agree with @keymon here even though it is more work. While this is probably safe, `insertbefore` and `line` both look potentially quite error-prone. I think `lineinfile` has caused us problems in the past.

mtekel, line 15: I have decided to use lineinfile as a nicer way of modifying the file otherwise (either by just appending or directly overwiritng). StatsD extension might go through more easily than graphite, which has now several PRs open for longer time...

annashipman:
This was discussed in Slack. @jimconner has found "an 'environment-variables service broker'... so you can do a tsuru service-bind stats -a appname to set the variables in a container"; @dcarley reckons it's this: https://github.com/tsuru/eviaas.

I think it's fine to leave it like for this PR and story, i.e. having `STATSD_HOST` and `STATSD_PORT` explicitly set per application. We might later want to take one of the approaches @keymon suggests or that @jimconner suggested, but it's not something we need to do for the prototype.

annashipman:
Also, a point on the wall for @keymon for referring to the platform user as "he"!

tsuru_graphite.yml
annashipman, line 15: We can just use a fork until it goes through, as Hector suggested. Not a great option if it takes a long time, but safer in this case.

dcarley:
For posterity: we closed this because creating a multi-tenant Graphite install is too complex for the prototype. We're going to test a third-party metrics-as-a-service.

PR #86, opened 2015-06-12
[#95695338] Shutdown insecure tsuru

keymon:
**What**

[#95695338 Shutdown insecure tsuru](https://www.pivotaltracker.com/n/projects/1275640/stories/95695338)

We remove the dependencies on the insecure plain HTTP api traffic on port 8080, before shutting down the LB routing in the terraform configuration. 

We've created this PR to be reviewed in the following context:
- We want to make tsuru API listen only on 127.0.0.1:8080, so it cannot be accessed from outside the server but only via the nginx SSL termination on port 443. 
- We want all the internal tsuru services connect via `HTTPS/443`  (Note: Currently only `gandalf` needs to be changed).

For this, we need to use new releases of these roles:
- `tsuru-api v0.0.3`  to [allow change listen parameters](https://github.com/alphagov/ansible-playbook-tsuru_api/pull/4)
- `gandalf v.0.0.2` to [allow change the remote tsuru API endpoint](https://github.com/alphagov/ansible-playbook-gandalf/pull/2)

Global variables to refer the `tsuru_api` internal and external LB and URL are also added.

**Who should review this PR**

Any one on the main team with the exception of @actionjack and @keymon

**Dependencies** 

This PR should be merged before the related one which deletes the HTTPS/8080 LB routing in `tsuru-terraform`, to avoid break the state.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -14,7 +14,8 @@
 - hosts: "{{ hosts_prefix }}-tsuru-gandalf*"
   sudo: yes
   roles:
-    - gandalf
+    - role: gandalf
+      tsuru_api_endpoint: "{{ tsuru_api_internal_url }}"
 
 - include: tsuru_api.yml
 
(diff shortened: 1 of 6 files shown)


jimconner:
I have tested this against my current environment on GCE and confirmed that 8080 comms are now blocked and 443 comms are working correctly. LGTM - merging.

PR #90, opened 2015-06-19
[#97198688] API SSL proxy disable buffering, increase read timeout

mtekel:
Add proxy_buffering off and change proxy_read_timeout to 15 minutes. This way any logs streamed via api server arrive at the client immediately (e.g. app logs, logs from adding platform etc.). Also, connection is not dropped quickly when you try to tail logs and there's no data incoming (previously that would happen after 15s).

### DIFF
--- a/api_ssl_proxy_vars.yml
+++ b/api_ssl_proxy_vars.yml
@@ -1,0 +1,25 @@
+---
+nginx_sites:
+  ssl_reverse_proxy:
+    - listen 443 ssl
+    - ssl_certificate wildcard.tsuru.paas.alphagov.co.uk.site.chain.crt
+    - ssl_certificate_key wildcard.tsuru.paas.alphagov.co.uk.key
+    - server_name {{ domain_name }}
+    - location / {
+      proxy_pass http://localhost:{{ upstream_port }}/;
+      proxy_redirect default;
+      proxy_redirect http://$host/ https://$host/;
+      proxy_redirect http://$hostname/ https://$host/;
+      proxy_read_timeout 900s;
+      proxy_connect_timeout 15s;
+      proxy_buffering off;
+      }
+  default:
+    - listen 80
+    - return 301 https://$host$request_uri
+nginx_configs:
+  proxy:
+    - proxy_set_header Host $http_host
+    - proxy_set_header X-Real-IP  $remote_addr
+    - proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for
+    - proxy_set_header X-Forwarded-Proto https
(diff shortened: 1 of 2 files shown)


dcarley:
Can you break this into two separate commits (and probably separate pull requests)?

We should try to make [commits atomic](http://seesparkbox.com/foundry/atomic_commits_with_git), so that they're easier to review and easier to rollback without affecting other functionality. A good indicator of when something should be broken apart is when I find myself writing the word "also" in a commit message.

Incidentally, I'm not sure about the `proxy_read_timeout` change. Won't it affect how quickly a failing backend is taken out of service and potentially leave lots of stale connections?

I was also unsure whether we should disable `proxy_buffering` for the router. Will a single application with many slow clients affect the performance/availability of other applications?

mtekel:
Regarding `proxy_buffering`: I'd leave it so. It's a reasonable default. As an app owner I don't expect my app output to be anyhow buffered or flow altered. If I want to do so, I can configure the web server my app is running in. Example is an app that streams some data to client, e.g. tsuru dashboard showing logs...

If we need better router performance in the future, we can tweak it then. There's also a lot of other options to improve performance (e.g. `keepalive` connections).

Regarding timeouts:
- if the hipache/api dies, TCP keepalive will fail, dropping the connection sooner
- google healthchecks don't seem to really work. If there's issue with hipache/api, nginx still accepts connections and returns 502. I think this requires some further config optimization.

But for now I can split it for routers and api...

mtekel:
Few more comments and findings:
- in my opinion a proxy or a load balancer should be completely transparent and not modify/obstruct flow in any way - unless specifically desired. E.g. ELB and google LBs don't buffer or anyhow alter flow.
- even with buffering disabled, there's some sort of buffering when doing reads. App ssh now opens, but when you type nothing happens. Even though you can see packets reaching nginx every time you press a button...

dcarley:
Can you simplify this PR to just deal with `proxy_buffering`? The `proxy_read_timeout` change isn't strictly related and discussing it is going to confuse/delay the other change.

As for whether we change buffering for both the API and Router, I don't mind. Technically it's of benefit to apps because it allows them to free up resources sooner. But they wouldn't have had it before and it makes our configs more complicated. Your call.

If we are going to split the configs and change it for just the API, there's no longer any benefit in having the variables in a separate file where they can be shared. We should also make the commits atomic - at the moment reverting a single commit would break it and it's not obvious from the diff what has changed from the config that it was copied from. It should be structured as:
1. Duplicate the file and change the reference to it
2. Modify a single variable

mtekel:
I can make this PR to deal with buffering only. But the timeout issue would still have to be solved on the API servers. It's not only that you get disconnected after 15s when you e.g. try to see logs. If platform addition doesn't produce any output for 15s, you get disconnected as well. Same for tsuru app-shell - or any other operations that have 15s gap in output...

mtekel:
Buffering PR continued here: https://github.com/alphagov/tsuru-ansible/pull/92

PR #91, opened 2015-06-22
[#97365070] Add new starter Richard Knop to vault

dcarley:
Added to recipients and re-encrypted.

(incidentally `make diff-vault` doesn't work in this scenario because
`vault_passphrase.gpg` is different between the two branches - oh well)


RichardKnop:
Thanks :+1:

PR #92, opened 2015-06-22
[#97198688] Disable nginx proxy buffering

mtekel:
Buffering is on by default on Nginx. It will cause it to wait for 1KB of output or close of connection before it sends reply back to original requestor. This affects situations where you want to stream response, e.g. when connecting to API server and adding platform or streaming logs. Disabling buffering makes Nginx behave more like ELB or google LBs in being more transparent and not interfering with the data flow from the apps.

More discussion [here](https://github.com/alphagov/tsuru-ansible/pull/90).

### DIFF
--- a/ssl_proxy_vars.yml
+++ b/ssl_proxy_vars.yml
@@ -12,6 +12,7 @@ nginx_sites:
       proxy_redirect http://$hostname/ https://$host/;
       proxy_read_timeout 15s;
       proxy_connect_timeout 15s;
+      proxy_buffering off;
       }
   default:
     - listen 80


dcarley:
You were going to test whether this works without modifying the read timeout. Does it?

mtekel:
Yes, this should just disable buffering. Timeout issue is being handled in [this](https://www.pivotaltracker.com/n/projects/1275640/stories/97521746) story.

keymon:
This works partly for me. 

Without buffering, I get immediate feedback while running commands like platform-add, etc.  

But there are still issues:
- `tsuru app-shell` fails with nginx, probably because it just uses the TCP connection of the HTTP request directly, without using something like websockets, so nginx causes problems here.
- `tsuru platform-add` still timeouts if there is no input/output for 15 seconds.

Anyway, from previous discussions we shouldn't mix the buffer disable and the timeout. I consider this is a step in the right direction, and we should merge it. 

For me :+1: 

@dcarley ?

dcarley:
Yep, it's a step in the right direction :+1:

keymon:
The  `app-shell` thing to use websockets is an issue in tsuru: https://github.com/tsuru/tsuru/issues/1162

I think app-logs should also use websockets.

PR #93, opened 2015-06-22
[#96562016] Create some basic smoke tests

saliceti:
This adds smoke testing capability to cover [Create some basic smoke tests](https://www.pivotaltracker.com/n/projects/1275640/stories/96562016)

Coverage:
-  login
-  add SSH key
-  create an application
-  create and bind a service
-  push application
-  connect to the application

Verify the smoke test can be run successfully (or not) against any deployed environment.

Follow the README for requirements and examples. Add verbose option to see what is going on in detail.

**How does the PR work?**
- We use [RSpec](http://rspec.info/). 
- The tests are defined in `spec/endtoend/endtoend_spec.rb`.
- **The tests are run in the order they are written** in the script since there is strong dependency between them. If one failed, the whole execution is aborted.
- RSpec configures and runs the `tsuru` client, then checks the output. We use ruby class wrappers around the _git_ and _tsuru_ commands to streamline the execution of external commands and capture the output in a standard way. Check `*_helper.rb` for more info.
- A `ssh-wrapper` script is created to allow us override the ssh config in a portable way.
- A temporary home directory is created for each test execution to guarantee a clean environment. 
- A new random application name is generated to avoid postgresql ROLE conflicts, because of [this bug](https://github.com/tsuru/postgres-api/issues/1).
- The _after(:all)_ hook cleans up the environment after each execution.

**Known issues**
1. The test will create a new role in postgres everytime it runs. This is caused by [this bug](https://github.com/tsuru/postgres-api/issues/1). We added an additional test to check this behaviour, which is marked as pending.
2. The current certificate in our tsuru platform is not valid for the default application names, so we must ignore the certificate verification. An additional pending test checks for a valid certificate.
3. After the application is deployed it is sometimes not immediately available and we receive error `502: Bad gateway`. The connection test that comes afterwards has a retry feature to work around this. The root cause will be investigated in [this story](https://www.pivotaltracker.com/n/projects/1275640/stories/97503258)

**About the PR history**

This code was originally in a different repository: https://github.com/alphagov/tsuru-testing. We merged the two repositories in this PR.

**Dependencies**
-  tsuru client
-  git
-  Ruby 2.2.2
-  Ideally [RVM](https://rvm.io/)

**Who should review this PR**
Anyone in the core team


dcarley:
Should `tsuru-testing/README.md` be removed because it duplicates the additions to the main `README.md`?

Should references to serverspec (`Gemfile`, `Gemfile.lock` and `spec/serverspec_helper.rb`) be omitted while we're not using it?

There are a lot of commits. Could any be squashed without losing information? I'm thinking of things like 2b8bf97768b9196badc85a5a2dddcf1787d0a04f where there is no functional change.

I haven't reviewed this fully, so not assigning..

actionjack:
Is it worth adding a Makefile target for the smoke tests (Make calling Rake, inception anyone?)

actionjack:
I've have been using this and it works but I won't be able merge it since I'm going to be out of office for the rest of this week.

keymon:
Yes, I guess it makes sense since we are using Makefile. We can add a `make test`, but we still will have to specify the tsuru user and password:

```
TSURU_USER=... TSURU_PASS=... make test-aws
TSURU_USER=... TSURU_PASS=... make test-gce
```

keymon:
I added logic to run the tests from `make`. 

As the credentials for tsuru are stored in the ansible vault, this commit uses `ansible-vault` command to automatically pick the  credentials.

In order to run the tests, you only need to do:

```
make test-aws DEPLOY_ENV=<env>
make test-gce DEPLOY_ENV=<env>
```

Makefile
```diff
@@ -58,6 +58,24 @@ recrypt: import-gpg-keys
 			$(shell cat gpg.recipients | awk -F: {'printf "-r "$$1" "'}) && \
 		ansible-vault encrypt ${VAULT_FILE}
 
+test-aws:
```
jimconner: Makefile usage string on line 17 for the all: case needs updating to show the newly available test-aws and test-gce command options - A minor point, but nice to have it shown in the usage string.

README.md
```diff
@@ -170,3 +170,53 @@ Use the `Makefile`. Run without any arguments for more information:
...
+
+```bash
+bundle install
+```
+
+#### Run
+
+You must pass the `TSURU_USER` and `TSURU_PASS` environment variables, and
+optionally the `DEPLOY_ENV` (defaults to `ci` if missing).
+
```
jimconner, line 204: Is is not necessary to specify TSURU_USER/PASS on as environment variables since you're pulling them out of the vault now... Docs should be updated.

spec/endtoend/endtoend_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      @tsuru_api_url = "https://#{RSpec.configuration.target_api_host}"
+      @tsuru_api_url_insecure = "http://#{RSpec.configuration.target_api_host}:8080"
+
+      @tsuru_command.target_add("ci", @tsuru_api_url)
+      @tsuru_command.target_add("ci-insecure", @tsuru_api_url_insecure)
+
+      @tsuru_user = RSpec.configuration.tsuru_user || raise("You must set 'TSURU_USER' env var")
+      @tsuru_pass = RSpec.configuration.tsuru_pass || raise("You must set 'TSURU_PASS' env var")
+
```
jimconner, line 24: Error message should be updated here to reflect that we're getting USER/PASS from the vault rather than as environment variables.

spec/endtoend/endtoend_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+        { :verbose => RSpec.configuration.verbose }
+      )
+      @git_command.clone("https://github.com/alphagov/flask-sqlalchemy-postgres-heroku-example.git")
+
+
+    end
+
+    after(:all) do
+      @tsuru_command.key_remove('rspec') # Remove previous state if needed
+      @tsuru_command.service_unbind('sampleapptestdb', @sampleapp_name)
```
jimconner: We are using 'sampleapp' + Time.now.to_i.to_s to ensure that our app name does not clash with any others on the system, but we've got the 'sampleapptestdb' db instance name hardcoded throughout - Is there a reason why we're not doing a unique name for the DB instance too?

spec/tsuru_helper.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  c.add_setting :target_platform, :default => ENV['TARGET_PLATFORM'] || 'aws'
+
+  # If nil will be populated based on the default platform
+  c.add_setting :target_api_host, :default =>
+    case c.target_platform
+    when 'aws'
+      "#{RSpec.configuration.deploy_env}-api.tsuru.paas.alphagov.co.uk"
+    when 'gce'
+      "#{RSpec.configuration.deploy_env}-api.tsuru2.paas.alphagov.co.uk"
```
jimconner: Is there a way we can take the domain_name parameter from platform-aws.yml and platform-gce.yml respectively rather than having them hardcoded in here?

jimconner:
I've tested this against an environment on GCE, and everything seems to work as expected. 10 examples, 0 failures and 2 pending due to known issues - Looks good to me apart from some documentation tweaks and wanting to get rid of some hardcoded values.

spec/endtoend/endtoend_spec.rb
keymon, line 24: I don't think the test code it-self should mention the vault. The test does not know how you call it, it is something external (Makefile)

spec/endtoend/endtoend_spec.rb
keymon: We use `'sampleapp' + Time.now.to_i.to_s` to workaround the issue in postgresapisql that it creates the role when binding but it does not delete it while unbind. The Role in postgres is somehow based on the name of the app. 

But you are right, if we randomise the db instance we can even run several instances of the test suite in parallel. 

I will do so.

spec/tsuru_helper.rb
keymon: That would couple the ansible code with the spec code :-m. Not sure if it is worth it.

spec/endtoend/endtoend_spec.rb
jimconner, line 24: Fair point - OK - I'm happy with that.

saliceti, line 24: Even if make is getting the values from the vault, for RSpec it is still an environment variable.

README.md
saliceti, line 204: Only if you use make. Updated the docs to make it clearer.

spec/tsuru_helper.rb
jimconner: Perhaps we could pass the domain across as an environment variable from the Makefile the same way that we're doing for TSURU_USER and TSURU_PASS - I would prefer not having the domain names configured in more than one place as it will undoubtedly cause confusion for us (or others) in the future.

spec/endtoend/endtoend_spec.rb
saliceti: In theory we always deleted what we create, so we could always use the same names. The unique app name is a workaround because of the postrgres api role bug.

keymon: I added the feature of pseudo-random for instances.

jimconner: Thanks - looks good.

jimconner:
All LGTM now - hitting the merge button.

PR #94, opened 2015-06-22
aws: Use IAM role/profile for registry on S3

dcarley:
The docker registry supports using IAM instance roles/profiles automagically
if you omit the access/secret key in the configuration. This was supported
by the upstream playbook in version 1.0.6:
- codingbunch/ansible-docker-registry/compare/v1.0.5...v1.0.6

This change to Terraform must be merged first:
- alphagov/tsuru-terraform#74

---

I tested this with `tsuru platform-add …`.


jimconner:
Tested on AWS - works perfectly. Merging now.

PR #96, opened 2015-06-23
Add --path vendor/bundle to install gem files

keymon:
We want the tests to run in a local environment, not install the packages globally. For that, we must pass the option `--path vendor/bundle`.


mtekel:
Works and fixes issue with jenkins trying to do sudo.

PR #97, opened 2015-06-23
Makefile does noise if it fails opening the vault

keymon:
We retrieve the Tsuru credentials from the ansible vault, but if it
fails getting it, it will not display the error because we capture the
stderr from the `ansible-vault` command in the Makefile.

In this commit we make the `open_the_vault.sh` less verbose to keep
output clean, and we stop capturing the stderr when calling
`ansible-vault`. 

This way, if it fails decrypting it will display the right error and we can 
troubleshoot quickly.

To test it, one can run:

```
GPG_AGENT_INFO=/tmp/test DEPLOY_ENV=ci make test-aws VERBOSE=1
```

That would make gpg fail, so you can see the behaviour with and without this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,14 +62,22 @@ recrypt: import-gpg-keys
 check-target-api-host-var:
 	@[ -z "${TARGET_API_HOST}" ] && echo TARGET_API_HOST cannot be empty && exit 2 || true
 
-test: check-target-api-host-var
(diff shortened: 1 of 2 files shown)


saliceti:
On my machine the GPG agent starts successfully automatically even with GPG_AGENT_INFO=/tmp/test. But I can see the GPG error messsage and if I don't enter my credentials the test doesn't run.
Merging.

PR #99, opened 2015-06-23
Fix minor issues in README related to smoketest

keymon:
`DEPLOY_ENV=...` now is mandatory when calling `make test-{gce,aws}` and new variable `TARGET_API_HOST`


mtekel:
More accurately corresponds with the current state of things.

PR #102, opened 2015-07-02
[#97503258] Revert "Added logic to retry the HTTPS connection test"

dcarley:
This reverts commit d1b2f792b80cebe8a4cced3bee8bd439d6bcccf9.

No longer seems to be necessary. I suspect that this was fixed by adding a
health check to the demo app, which prevents a route from being added until
the app/container has been confirmed started/working:
- alphagov/flask-sqlalchemy-postgres-heroku-example#4

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ gem 'rspec'
 gem 'serverspec', '~> 2.18.0'
 gem 'ansible_spec'
 gem 'highline'
-gem 'rspec-retry'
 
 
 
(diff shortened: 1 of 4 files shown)


saliceti:
What about if another app doesn't have a healthcheck configured?
I would prefer to test both cases. Or maybe we keep this for the integration tests, as you prefer.

dcarley:
The retry logic was only added to workaround undesired behaviour that couldn't be fixed at the time. Now that we know how to fix it, we should no longer allow that undesired behaviour and we should be made aware if there's a regression that causes it to happen again (even if it's caused by us changing the app we use to test).

I don't think there's value in having a separate test for it. What would it tell us?

saliceti:
We agreed with @dcarley that all apps must have healthcheck configured, so we should focus on testing this case.
Merging

PR #106, opened 2015-07-06
[#97818382] Add tests for Tsuru API healthchecks

dcarley:
To test that all of the components (MongoDB, Router, Registry, etc) are
passing health checks by using this endpoint:
- http://docs.tsuru.io/en/master/reference/api.html#full-healthcheck-of-all-tsuru-components

Technically we don't need to check the output because the endpoint should
return a non-200 response if any of the components are not `WORKING`.
However for completeness I am checking that that we have at least three
lines of output so that the `each` loop gets called and that each line
contains `WORKING` in roughly the right place.

I've pulled some variables that I need into the outer `describe` block so
that I can use them across both contexts. I've also referenced
`@tsuru_api_host`, since we're setting it, rather than going back to
`RSpec.configuration.target_api_host` each time.
#### Use certifi CA bundle for end-to-end tests

Otherwise `open-uri` raises a certificate verify exception for the GlobalSign
certificate that we have on the API when running the tests from Mac OS X:

```
Failures:

  1) TsuruEndToEnd tsuru API should pass healthchecks for all components
     Failure/Error: response = URI.parse("#{@tsuru_api_url}/healthcheck/?check=all").open()
     OpenSSL::SSL::SSLError:
       SSL_connect returned=1 errno=0 state=unknown state: certificate verify failed
     # ./spec/endtoend/endtoend_spec.rb:15:in `block (3 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.2.0/gems/rspec-retry-0.4.0/lib/rspec/retry.rb:43:in `block (3 levels) in apply'
     # ./vendor/bundle/ruby/2.2.0/gems/rspec-retry-0.4.0/lib/rspec/retry.rb:34:in `times'
     # ./vendor/bundle/ruby/2.2.0/gems/rspec-retry-0.4.0/lib/rspec/retry.rb:34:in `block (2 levels) in apply'
```

We could install the Ruby version of `certifi` certificate path, however it
seems to require monkey-patching to tell `open-uri` to always use that CA path.
Using the environment variable is easier and at least consistent with how we
run Ansible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,8 @@ check-target-api-host-var:
 
 test: check-env-var check-target-api-host-var load-tsuru-creds
 	@bundle install --path vendor/bundle --quiet
-	@TARGET_API_HOST=${TARGET_API_HOST} \
+	@SSL_CERT_FILE=$(shell python -m certifi) \
+	TARGET_API_HOST=${TARGET_API_HOST} \
 	TSURU_USER=${TSURU_USER} \
 	TSURU_PASS=${TSURU_PASS} \
 	bundle exec rake endtoend:all
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -1,3 +1,4 @@
+require 'net/http'
 require 'open-uri'
 require 'openssl'
 require 'git_helper'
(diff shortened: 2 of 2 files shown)


mtekel:
Why only 3 lines? Don't we want to check that _all_ components work? 

```
MongoDB: WORKING (1.052855ms)
Router Hipache: WORKING (1.74112ms)
docker-registry: WORKING (259.669203ms)
Gandalf: WORKING (2.935942ms)
```

mtekel:
Tested with number raised to 4. After I stopped gandalf, check returns 500 already, so the count and individual health of services is never checked. Nevertheless that information could be useful (about which services work and which not). Perhaps split this into 2 tests?

dcarley:
> Why only 3 lines? Don't we want to check that all components work?

I should have said in the commit. We only need to check that we have enough that it looks like the right output, so I picked the current number minus one (which would allow us e.g. to switch the router to something that doesn't implement health checks). The test will always check all of the lines, but it will fail if there are less than three.

dcarley:
On reflection, checking the response body might not provide any value at all. Unless we're able to report the particular component that failed. Maybe we should just remove that part and rely on the status code?

mtekel:
I think we can have 2 tests:
- one that reports specific component not working first
- and the second test general non-200 response. Which would indicate that all components are `WORKING`, but still something is wrong

dcarley:
Rebased with separate tests and another fix.

PR #107, opened 2015-07-06
[#98450546] Set proxy timeout to default

mtekel:
For some reason, we had proxy_read_timeout set to 15s. This was too short for some operations and causing problems when:
- tailing logs and there was no events for more thatn 15s [#97521746](https://www.pivotaltracker.com/n/projects/1275640/stories/97521746)
- deploying app and it took more than 15s for API to clone the repo [#98450546](https://www.pivotaltracker.com/n/projects/1275640/stories/98450546)
- deploying app and it took more than 15s when uploading final package to registry [#97521746](https://www.pivotaltracker.com/n/projects/1275640/stories/97521746)

Hence reverting proxy_read_timeout setting to default, which is 60s. This fixes the issues described in the above tracker stories.

### DIFF
--- a/ssl_proxy_vars.yml
+++ b/ssl_proxy_vars.yml
@@ -10,7 +10,6 @@ nginx_sites:
       proxy_redirect default;
       proxy_redirect http://$host/ https://$host/;
       proxy_redirect http://$hostname/ https://$host/;
-      proxy_read_timeout 15s;
       proxy_connect_timeout 15s;
       proxy_buffering off;
       }


dcarley:
> deploying app and it took more than 15s for API to clone the repo #98450546
> deploying app and it took more than 15s when uploading final package to registry #97521746

Is it possible to add a test for either of these to prevent regressions?

saliceti:
The default is **60s**, but I would assume for worse case we can easily reach that limit (big app, slow connection...). We could set it to **5 min**?
We may also hit the same issue with write operations, so we could set _proxy_send_timeout_ to the same time.

mtekel:
@dcarley I have asked Brett to create new tracker stories to cover addition of tests.

annashipman:
I think Brett has already spoken to Michal offline about this, but for those following along: the tests are always part of the same story and should be done alongside the work, not in a separate story. OK for them to be in a separate PR if necessary, but not a separate story.

mtekel:
The catch here is that this is a very low level configuration setting that impacts HTTP communication with API. This impacts ALL platform actions that need to talk to API and have some gaps in communication (usually where API is doing something or waiting on st. else - e.g. app logs). It is impossible now to write a test for every functionality that uses API. The test for tailing logs is here: https://github.com/alphagov/tsuru-ansible/pull/109

I have tried to find and app that takes long to upload/clone, approximately consistently close to say 55s, but it's difficult. I would have to create and artificial repository with thousands of files to ensure cloning is long. The smoketests then would of course take that much time more. Regarding long uploads: There's another issue, with nginx checking and preventing any objects bigger than 1MB being sent. The fix is again 1 line in config, tests can be much more complicated, slow and artificial.

mtekel:
See https://github.com/alphagov/tsuru-ansible/pull/110

dcarley:
I've changed the subject to reference the other story which is the one on the board and better describes the immediate problem with the demo environment. Merging now, but we shouldn't close that story until we've resolved how to test it.

PR #109, opened 2015-07-07
[#97521746] Timeout smoketest

mtekel:
read_proxy_timeout default is 60s. When doing app-log -f and no data is transferred, you should not be disconnected sooner than in this period. The added smoketest verifies this.

### DIFF
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -103,20 +103,20 @@ describe "TsuruEndToEnd" do
       expect(@tsuru_command.stdout).to match /Instance .* is now bound to the app .*/
     end
 
-    it "Should be able to push the application" do
+    it "should be able to push the application" do
       git_url = @tsuru_command.get_app_repository(@sampleapp_name)
       expect(git_url).not_to be_nil
       @git_command.push(git_url)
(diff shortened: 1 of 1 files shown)


dcarley:
This will extend the time it takes to run the smoke tests by 1 minute. Is it that valuable to test this compared to the other bugs described in https://github.com/alphagov/tsuru-ansible/pull/107#issuecomment-118875032?

mtekel:
To test timeout... you need to wait for it to happen. To test that 504 when deploying app, I would need to find and app that takes consistently more than 15s to clone the repository. Ideally one that takes 50s, so that the test is good. I am not sure how to achieve that consistently. Then I can reasonably easily test if the deploy happened or if it resulted in 504...

annashipman:
We've discussed this offline, and @keymon and @mtekel are working on creating a test that recreates one of the actual bugs manifested by this config change, which will be a more useful test for us. We care about behaviour, not the precise configuration of nginx.

mtekel:
@annashipman but this is test specifically verifying that log tail cutoff as described in #97521746 doesn't happen again. This is an actual bug.

annashipman:
@mtekel as you know, because we have discussed this several times offline, and because Dan has commented above, this test is slow. This PR is closed.

PR #110, opened 2015-07-07
Remove http upload object limit

mtekel:
NGINX has default client_max_body_size set to 1MB. If you try to deploy app using tsuru app-deploy and there's any object larger than 1 MB (e.g. any media, pictures, binaries, packages), you get a 413 error. This commit removes the limit checking, removing another point of proxy interference with the platform/client communication.

### DIFF
--- a/ssl_proxy_vars.yml
+++ b/ssl_proxy_vars.yml
@@ -11,6 +11,7 @@ nginx_sites:
       proxy_redirect http://$host/ https://$host/;
       proxy_redirect http://$hostname/ https://$host/;
       proxy_read_timeout 15s;
+      client_max_body_size 0;
       proxy_connect_timeout 15s;
       proxy_buffering off;
       }


keymon:
:+1: it does work. I was able to reproduce the issue and fix it with this patch. 

But @dcarley suggests that we should discuss what is the sensible limit for this parameter, as we shouldn't allow any size upload.  Makes sense, as it is a error-prone process and a user might upload the all their `$HOME` with the last holidays videos by mistake :)

I add the instructions to help testing the PR:

# How I tested this PR?
1. Create a big file in the [flask app](https://github.com/alphagov/flask-sqlalchemy-postgres-heroku-example): 
   
   ```
   git clone https://github.com/alphagov/flask-sqlalchemy-postgres-heroku-example
   cd flask-sqlalchemy-postgres-heroku-example
   dd if=/dev/urandom of=big_file bs=1k count=2000 # 2M file
   ```
2. Try to push this via `tsuru app-deploy` in an environment without this option. It fails:
   
   ```
   tsuru app-create test python
   tsuru service-add postgresql testdb
   tsuru service-bind  testdb -a test
   # deploy the app
   tsuru app-deploy -a test .
   ```
   
   Fails with: 
   
   ```
   $ tsuru app-deploy -a test .
   Uploading files...Error: <html>
   <head><title>413 Request Entity Too Large</title></head>
   <body bgcolor="white">
   <center><h1>413 Request Entity Too Large</h1></center>
   <hr><center>nginx</center>
   </body>
   </html>
   ```

Now, with the change:
1. Apply this change:
   
   ```
   cd tsuru-ansible
   git fetch && git checkout fix-upload-limit
   make gce DEPLOY_ENV=hector070715 ARGS='-l *tsuru-api*'
   ```
2. And repeat the test, now works :):
   
   ```
   $ cd flask-sqlalchemy-postgres-heroku-example
   $ tsuru app-deploy -a test .
   Uploading files..... ok
   Reading package lists...
   Building dependency tree...
   (...)
   ```

dcarley:
Can you talk to Brett about adding a new story/bug for this?

We should use that to discuss how it will affect users (did this come up when deploying Marketplace or was it an accident when deploying from the wrong directory?), what a sensible limit (if anything) is, whether we can test it, and what priority it is to fix.

brettansley:
For issues like this it would be really useful to understand how it was discovered - which app were you deploying when you first found the limit? This helps to identify what first caused this so we can better define a solution.

PR #113, opened 2015-07-09
[#98109110] Update ssh template to use ubuntu NAT user on AWS

mtekel:
AWS NAT is no longer custom AWS linux box, but runs instead on ubuntu like all other hosts and GCE NAT. Update ssh template to reflect these changes.

Depends on https://github.com/alphagov/tsuru-terraform/pull/80


keymon:
The team members might need to re-provision their nat servers if you want to keep using a running environment after these PR are merged. The reviewer should let the team know.

```
terraform taint aws_instance.nat && terraform apply
```

PR #114, opened 2015-07-09
[#98109110] Suspend and resume environments via make

mtekel:
Adding functionality to suspend and resume environments:
- enhance gce.py script to be able to start instances
- enhance ec2.py script to list only stopped instances
- add ansible playbook to start all VMs in a specified environment on EC2
- add ansible playbook to start all VMs in a specified environment on GCE
- add makefile targets to call suspend and resume actions on AWS and GCE

Depends on https://github.com/alphagov/tsuru-ansible/pull/113

### DIFF
--- a/gce-wake.yml
+++ b/gce-wake.yml
@@ -1,0 +1,15 @@
+---
+
+# Start all the instances in the specified GCE deployment
+  - hosts: localhost
+    tasks:
+      - name: Deploy gce.py
+        file: src=gce.py dest=./gce.py mode=0744
+      - name: Deploy secrets.py
+        file: src=secrets.py dest=./secrets.py mode=0700
+
+      - name: Wake instances
+        shell: "./gce.py --start {{ item }}"
+        when: '"{{ deploy_env }}-tsuru-" in item'
+        with_items: groups['status_terminated']
+
--- a/gce-wake.yml
+++ b/gce-wake.yml
@@ -10,6 +10,6 @@
 
       - name: Wake instances
         shell: "./gce.py --start {{ item }}"
-        when: '"{{ deploy_env }}-tsuru-" in item'
+        when: item | match("^{{ deploy_env }}-*")
         with_items: groups['status_terminated']
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ ssh.config
 .ansible-galaxy.check
 .diff-vault.fifo
 vendor/
+ansible-ec2.cache
+ansible-ec2.index
--- a/Makefile
+++ b/Makefile
@@ -103,11 +103,11 @@ start-gce: check-env-var render-ssh-config
 	SSL_CERT_FILE=$(shell python -m certifi) ansible-playbook -i gce.py gce-wake.yml -e deploy_env=${DEPLOY_ENV}
 
 suspend-aws: check-env-var render-ssh-config
-	ansible all -i ec2.py -a 'sudo poweroff' -l "!tag_Name_${DEPLOY_ENV}-tsuru-nat:tag_Name_${DEPLOY_ENV}-tsuru-*"
-	ansible all -i ec2.py -a 'sudo poweroff' -l "tag_Name_${DEPLOY_ENV}-tsuru-nat"
+	ansible all -i ec2.py -a 'sudo poweroff' -l "!~^tag_Name_${DEPLOY_ENV}-tsuru-nat:~^tag_Name_${DEPLOY_ENV}-"
+	ansible all -i ec2.py -a 'sudo poweroff' -l "~^tag_Name_${DEPLOY_ENV}-tsuru-nat"
 
 suspend-gce: check-env-var render-ssh-config
 	SSL_CERT_FILE=$(shell python -m certifi) \
-	ansible all -i gce.py -a 'sudo poweroff' -l "!${DEPLOY_ENV}-tsuru-nat:${DEPLOY_ENV}-tsuru-*"
+	ansible all -i gce.py -a 'sudo poweroff' -l "!~^${DEPLOY_ENV}-tsuru-nat:~^${DEPLOY_ENV}-"
 	SSL_CERT_FILE=$(shell python -m certifi) \
-	ansible all -i gce.py -a 'sudo poweroff' -l "${DEPLOY_ENV}-tsuru-nat"
+	ansible all -i gce.py -a 'sudo poweroff' -l "~^${DEPLOY_ENV}-tsuru-nat"
--- a/Makefile
+++ b/Makefile
@@ -100,12 +100,14 @@ start-aws: check-env-var render-ssh-config
 	ansible-playbook -i ec2.py ec2-wake.yml -e deploy_env=${DEPLOY_ENV}
 
 start-gce: check-env-var render-ssh-config
-	ansible-playbook -i gce.py gce-wake.yml -e deploy_env=${DEPLOY_ENV}
+	SSL_CERT_FILE=$(shell python -m certifi) ansible-playbook -i gce.py gce-wake.yml -e deploy_env=${DEPLOY_ENV}
 
 suspend-aws: check-env-var render-ssh-config
 	ansible all -i ec2.py -a 'sudo poweroff' -l "!tag_Name_${DEPLOY_ENV}-tsuru-nat:tag_Name_${DEPLOY_ENV}-tsuru-*"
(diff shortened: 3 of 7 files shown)


gce-wake.yml
```diff
@@ -0,0 +1,17 @@
+---
+
+# Start all the instances in the specified GCE deployment
+  - hosts: localhost
+    tasks:
+      - name: Deploy gce.py
+        file: src=gce.py dest=./gce.py mode=0744
+      - name: Deploy gce_account.pem
+        file: src=gce_account.pem dest=./gce_account.pem mode=0600
```
dcarley: Not everybody uses the PEM file (df316d612978d1a0009fa0f1ba880b6cabd75a6b) and they probably won't have it in this exact location (mine is `~/projects/paas/creds/gce-account.json`).

Can we parse `secrets.py` (which we know the location of) to get the path from `GCE_PARAMS`?

dcarley: Although I might have misunderstood this task. It's copying to and from `localhost`?

mtekel: I'm copying from localhost (current directory - thus ansible-terraform dir) to the localhost temp directory where the script runs (/tmp/ansible23423433). Otherwise I'd have to specify full path. 
Also, given that the full path is already in secrets.py, I can probably remove copying pem/json altogether...

dcarley: Gotcha. That sounds reasonable. If anyone is using an unqualified path in their config they should speak up now.

mtekel:
Updated...

dcarley:
This didn't stop the new `${DEPLOY_ENV}-influx-grafana` machines added in alphagov/tsuru-terraform/pull/81 because the VM names don't match `${DEPLOY_ENV}-tsuru-*`. We can either:
- rename the VMs to match the same convention
- change the `--limit` patterns to match `${DEPLOY_ENV}-*`

It seems like the latter might be safer and prevent similar mistakes in the future. Although we might need to be careful that we anchor it to the beginning of the string. What do you think?

mtekel:
`${DEPLOY_ENV}-*` pattern is not so safe. If you make some typo/mistake and set e.g. `DEPLOY_ENV=a`, then it will match all servers which have `a-` in name. Which could be more than what you have intended. I have specifically selected the `-tsuru-` variant because it is more safe. We don't tag hosts in any other way, so the matching is only by name...

dcarley:
I had to add this to get it running on Mac:

``` diff
diff --git a/Makefile b/Makefile
index d79a2c0..59c903a 100644
--- a/Makefile
+++ b/Makefile
@@ -100,12 +100,12 @@ start-aws: check-env-var render-ssh-config
        ansible-playbook -i ec2.py ec2-wake.yml -e deploy_env=${DEPLOY_ENV}

 start-gce: check-env-var render-ssh-config
-       ansible-playbook -i gce.py gce-wake.yml -e deploy_env=${DEPLOY_ENV}
+       SSL_CERT_FILE=$(shell python -m certifi) ansible-playbook -i gce.py gce-wake.yml -e deploy_env=${DEPLOY_ENV}

 suspend-aws: check-env-var render-ssh-config
        ansible all -i ec2.py -a 'sudo poweroff' -l "!tag_Name_${DEPLOY_ENV}-tsuru-nat:tag_Name_${DEPLOY_ENV}-tsuru-*"
        ansible all -i ec2.py -a 'sudo poweroff' -l "tag_Name_${DEPLOY_ENV}-tsuru-nat"

 suspend-gce: check-env-var render-ssh-config
-       ansible all -i gce.py -a 'sudo poweroff' -l "!${DEPLOY_ENV}-tsuru-nat:${DEPLOY_ENV}-tsuru-*"
-       ansible all -i gce.py -a 'sudo poweroff' -l "${DEPLOY_ENV}-tsuru-nat"
+       SSL_CERT_FILE=$(shell python -m certifi) ansible all -i gce.py -a 'sudo poweroff' -l "!${DEPLOY_ENV}-tsuru-nat:${DEPLOY_ENV}-tsuru-*"
+       SSL_CERT_FILE=$(shell python -m certifi) ansible all -i gce.py -a 'sudo poweroff' -l "${DEPLOY_ENV}-tsuru-nat"
```

Otherwise I get:

```
(ansible)➜  tsuru-ansible git:(env-suspend) make suspend-gce DEPLOY_ENV=dcarley
sed "s/DEPLOY_ENV/dcarley/g" ssh.config.template > ssh.config
ansible all -i gce.py -a 'sudo poweroff' -l "!dcarley-tsuru-nat:dcarley-tsuru-*"
…
RuntimeError: No CA Certificates were found in CA_CERTS_PATH. For information on how to get required certificate files, please visit https://libcloud.readthedocs.org/en/latest/other/ss
l-certificate-validation.html

make: *** [suspend-gce] Error 1
```

dcarley:
> If you make some typo/mistake and set e.g. DEPLOY_ENV=a, then it will match all servers which have a- in name.

Wouldn't anchoring the beginning of the string prevent that from happening?

mtekel:
I was looking at jinja2 [documentation](http://jinja.pocoo.org/docs/dev/templates/#builtin-filters), but could not find anything that would match regexp, or st. like starts-with. But in ansible jinja2 [documentation](https://docs.ansible.com/playbooks_filters.html) they mention some more methods that could do that. I will test it now. I will also check if/how you can use proper regexp for `--limit`.

keymon:
A example of this issue can be two environments: `test-tsuru-api` and `test-myfeature-tsuru-api`. The first `test-*` matches the second.

Anyway, is that such a big deal? We can assume the risk and be careful of not using a name like 'test-*'

mtekel:
Updated makefile. What's the decision for naming? Shall we require `-tsuru-` or do we just stick with `^{DEPLOY_ENV}-*` ?

dcarley:
Both are susceptible to mistakes. We could probably sanity check environment naming to prevent hyphens (and underscores, as Jim noted today). Whereas we can't enforce the naming of resources in alphagov/tsuru-ansible, so I'm inclined to think we should change the pattern.

dcarley:
I think we may need to invalidate the cache of `ec2.py` after fetching stopped instances. I just experienced a case where the first run of `make start-aws` seemed to miss one instance (which I can live with and I'm not sure we can investigate now). So I ran `make start-aws` again and then ran Ansible to show which hosts I had, but it only found the most recently started:

```
(ansible)➜  tsuru-ansible git:(env-suspend) ✗ make aws DEPLOY_ENV=dcarley ARGS=--list-hosts
sed "s/DEPLOY_ENV/dcarley/g" ssh.config.template > ssh.config
ansible-playbook -i ec2.py site-aws.yml -e "@platform-aws.yml" -e "deploy_env=dcarley" --list-hosts

playbook: site-aws.yml

  play #1 (tag_Name_dcarley-tsuru-registry*): host count=0

  play #2 (tag_Name_dcarley-tsuru-db): host count=0

  play #3 (tag_Name_dcarley-tsuru-db): host count=0

  play #4 (tag_Name_dcarley-tsuru-gandalf*): host count=0

  play #5 (tag_Name_dcarley-tsuru-api*): host count=1
    10.128.13.238

  play #6 (tag_Name_dcarley-tsuru-gandalf*): host count=0

  play #7 (tag_Name_dcarley-tsuru-docker*): host count=0

  play #8 (tag_Name_dcarley-tsuru-postgres): host count=0

  play #9 (tag_Name_dcarley-tsuru-router*): host count=0

  play #10 (tag_Name_dcarley-influx-grafana*): host count=0

  play #11 (tag_Name_dcarley-*): host count=1
    10.128.13.238

  play #12 (tag_Name_dcarley-tsuru-api*[0]): host count=1
    10.128.13.238
```

I had to run `/ec2.py --refresh-cache >/dev/null` before it worked again as expected.

mtekel:
start-aws calls ec2-wake playbook which explicitly does --refresh-cache. However it seems it takes a while (min or two) to notice that the servers are stopped. I think it has a bit to do with AWS itself not changing state immediately...

dcarley:
> start-aws calls ec2-wake playbook which explicitly does --refresh-cache. However it seems it takes a while (min or two) to notice that the servers are stopped. I think it has a bit to do with AWS itself not changing state immediately...

I think we need to call `ec2.py --refresh-cache` without `--stopped` afterwards otherwise the cache only contains (previously) stopped instances.

dcarley:
Outstanding:
- cache issue
- stopping influx instance

Otherwise this is :+1: :+1: :+1: to go! Tested on AWS and GCE several times.

mtekel:
updated...

dcarley:
Successfully brought up my AWS environment again this morning. Thanks @mtekel @keymon !

PR #117, opened 2015-07-13
[#98748824] Bump and pin docker server

dcarley:
#### Bump docker_server playbook to v0.1.0

Which includes support for:
- pinning the package version
- binding the default UNIX socket in addition to TCP
#### Pin docker server to 1.7.0

To prevent unintended and breaking upgrades whenever `get.docker.com` make a
new package available.


mtekel:
Works, correct versions are being installed - and they don't break anything.

PR #119, opened 2015-07-14
Add support to smoke tests for admin user being in multiple teams

jimconner:
**What**

When a Tsuru user is a member of multiple teams, they must specify at the time of app-create which team that app belongs to by using the `-t <teamname>` parameter on the end of the command.
Our demo environment has the admin user belonging to multiple teams, so the smoke tests are currently failing as app-create is not specifying a team.

This PR adds the required `-t admin` to the end of the app-create helper function:

**How to review this PR**

I've created the PR to be reviewed in the following context:
- I need smoke tests to correctly deploy the sampleapp application when the admin user belongs to multiple teams.

**Who should review this PR**
- Anyone in the core team

### DIFF
--- a/spec/tsuru_helper.rb
+++ b/spec/tsuru_helper.rb
@@ -67,7 +67,7 @@ class TsuruCommandLine < CommandLineHelper
   end
 
   def service_add(service_name, service_instance_name, plan)
-    execute_helper('tsuru', 'service-add', service_name, service_instance_name, plan)
+    execute_helper('tsuru', 'service-add', service_name, service_instance_name, plan, '-t', 'admin')
   end
 
   def service_remove(service_instance_name)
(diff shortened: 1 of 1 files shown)


dcarley:
It got a bit further, but also needs adding to `service-bind`:

```
Executing: HOME='/var/folders/nv/8xk4b8g904b3hnvys5zrz3pw0000gn/T/tsuru-command20150714-94689-1vr165f' tsuru service-add postgresql tst7887886341 shared
stderr: Error: user is member of multiple teams, please specify the team that owns the service instance
Exit code: 1
    should be able to create a service (FAILED - 1)
```

Should we assert that the user running the smoke tests belongs to more than one team?

dcarley:
Confirmed working when I was a member of two teams. Postponing discussion about what extra things the tests should setup/assert for now.

PR #120, opened 2015-07-14
[#98109110] Makefile: add exit codes on env. suspend

mtekel:
Because environment suspend target in Makefile has 2 commands and make only returns exit code of the last command, we only see exit code of the last command. We want to know if the first command succeded as well. Therefore remember exit codes of both commands and return sum of them - which is 0 in case both worked.

This is needed for the jenkins environment shutdown [job](https://github.com/alphagov/multicloud-deploy/pull/34), so that we can know if any of the shutdown commands inside Makefile have failed.


dcarley:
I changed the commands in order to test this.

With the first command failing and the second passing:

``` diff
diff --git a/Makefile b/Makefile
index 56d738b..fcc4ebe 100644
--- a/Makefile
+++ b/Makefile
@@ -103,9 +103,9 @@ start-gce: check-env-var render-ssh-config
        SSL_CERT_FILE=$(shell python -m certifi) ansible-playbook -i gce.py gce-wake.yml -e deploy_env=${DEPLOY_ENV}

 suspend-aws: check-env-var render-ssh-config
-       ansible all -i ec2.py -a 'sudo poweroff' -l "!~^tag_Name_${DEPLOY_ENV}-tsuru-nat:~^tag_Name_${DEPLOY_ENV}-"
+       false
        result=$?
-       ansible all -i ec2.py -a 'sudo poweroff' -l "~^tag_Name_${DEPLOY_ENV}-tsuru-nat"
+       true
        exit `expr $? + $result`

 suspend-gce: check-env-var render-ssh-config
```

Only the first command gets run:

```
➜  tsuru-ansible git:(make-exitcodes) ✗ make suspend-aws DEPLOY_ENV=foo
sed "s/DEPLOY_ENV/foo/g" ssh.config.template > ssh.config
false
make: *** [suspend-aws] Error 1
```

With the first command passing and the second failing:

``` diff
diff --git a/Makefile b/Makefile
index 56d738b..a10f3c0 100644
--- a/Makefile
+++ b/Makefile
@@ -103,9 +103,9 @@ start-gce: check-env-var render-ssh-config
        SSL_CERT_FILE=$(shell python -m certifi) ansible-playbook -i gce.py gce-wake.yml -e deploy_env=${DEPLOY_ENV}

 suspend-aws: check-env-var render-ssh-config
-       ansible all -i ec2.py -a 'sudo poweroff' -l "!~^tag_Name_${DEPLOY_ENV}-tsuru-nat:~^tag_Name_${DEPLOY_ENV}-"
+       true
        result=$?
-       ansible all -i ec2.py -a 'sudo poweroff' -l "~^tag_Name_${DEPLOY_ENV}-tsuru-nat"
+       false
        exit `expr $? + $result`

 suspend-gce: check-env-var render-ssh-config
```

It tries to execute the expansion of the `$?` internal Make variable:

```
➜  tsuru-ansible git:(make-exitcodes) ✗ make suspend-aws DEPLOY_ENV=foo
sed "s/DEPLOY_ENV/foo/g" ssh.config.template > ssh.config
true
result=check-env-var render-ssh-config
/bin/sh: render-ssh-config: command not found
make: *** [suspend-aws] Error 127
```

mtekel:
After a short discussion we'll be not doing this at all. The makefile lines are separate shells and make will fail already when the 1st command fails - which is what we want. The NAT shutdown will not be attempted, which lets us re-run the scripts or debug why shutdown failed.

PR #128, opened 2015-07-20
[#95168192] Unify AWS and GCE node names 

mtekel:
There is no reason why we should have different node names between AWS and GCE. Having same node names makes our ansible and terraform configuration more clear and straightforward.

Depends on https://github.com/alphagov/tsuru-terraform/pull/85

### DIFF
--- a/platform-aws.yml
+++ b/platform-aws.yml
@@ -8,7 +8,7 @@ ssl_key: "{{ aws_ssl_key }}"
 redis_host_name: "{{ hosts_prefix }}-tsuru-db"
 redis_host: "{{ hostvars[groups[redis_host_name][0]][ip_field_name] }}"
 
-tsuru_api_host_name: "{{ hosts_prefix }}-tsuru-api"
+tsuru_api_host_name: "{{ hosts_prefix }}-tsuru-api-0"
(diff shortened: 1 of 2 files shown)


keymon:
I had problems here. The register of the docker nodes was hanging, asking for a password running `tsuru-admin` in the task `Register node`

There were some problems:
- First [we run `tsuru-admin` commands without closing the stdin](https://github.com/alphagov/tsuru-ansible/blob/master/common.yml#L50), which can potentially block the process asking for a password. We should add `< /dev/null` in each line to prevent that.
- Second, the authentication [happens only in one random node of the api](https://github.com/alphagov/tsuru-ansible/blob/master/tsuru_api.yml#L49) because `run_once: true`. But this is state that is stored in the API server, so should be executed in both servers.
- Finally, in the [common.yml when registering the node we force it to use the first host](https://github.com/alphagov/tsuru-ansible/blob/master/common.yml) which maybe is the one which authenticated before, or maybe not.

annashipman:
As per https://github.com/alphagov/tsuru-terraform/pull/85, work should be related to a story. Which one is this related to? Can you update the PR title please?

mtekel:
Updated.

PR #127, opened 2015-07-17
[#98371938] Add Postgres standby server

mtekel:
Configure second postgres node as a standby and setup replication from master. Ensure playbook doesn't take too much time by running configuration of both master and standby in parallel. Provide easy way to promote master by defining master server in a variable.

Depends on https://github.com/alphagov/tsuru-terraform/pull/84
### How to test:

After doing ansible run, check that:
1 . initial backup exists, on master or slave do `sudo /usr/bin/envdir /etc/wal-e/ wal-e backup-list`, you should get st. like

```
wal_e.main   INFO     MSG: starting WAL-E
        DETAIL: The subcommand is "backup-list".
        STRUCTURED: time=2015-07-17T15:59:04.376411-00 pid=15043
name    last_modified   expanded_size_bytes wal_segment_backup_start    wal_segment_offset_backup_start wal_segment_backup_stop wal_segment_offset_backup_stop
base_000000010000000000000002_00000040  2015-07-16T14:38:52.484Z        000000010000000000000002    00000040
```

2 . Check postgres log on master: `sudo vim /var/log/postgresql/postgresql-9.3-main.log`, should contain st. like

```
wal_e.main   INFO     MSG: starting WAL-E
        DETAIL: The subcommand is "wal-push".
        STRUCTURED: time=2015-07-17T15:33:59.839938-00 pid=15056
wal_e.worker.upload INFO     MSG: begin archiving a file
        DETAIL: Uploading "pg_xlog/000000010000000000000006" to "s3://russ-mcp-postgres-backup/wal-e/wal_005/000000010000000000000006.lzo".
        STRUCTURED: time=2015-07-17T15:33:59.885894-00 pid=15056 action=push-wal key=s3://russ-mcp-postgres-backup/wal-e/wal_005/000000010000000000000006.lzo prefix=wal-e/ seg=000000010000000000000006 state=begin
wal_e.worker.upload INFO     MSG: completed archiving to a file
        DETAIL: Archiving to "s3://russ-mcp-postgres-backup/wal-e/wal_005/000000010000000000000006.lzo" complete at 61.661KiB/s.
        STRUCTURED: time=2015-07-17T15:34:01.308309-00 pid=15056 action=push-wal key=s3://russ-mcp-postgres-backup/wal-e/wal_005/000000010000000000000006.lzo prefix=wal-e/ rate=61.661 seg=000000010000000000000006 state=complete
wal_e.main   INFO     MSG: starting WAL-E
[...]

PR #129, opened 2015-07-20
[#95168192] Common registry playbook for GCE and AWS

mtekel:
Create unified registry role that can run on both GCE and AWS, so that we don't have to have two different site-<aws|gce> playbooks. Rename common.yml to site.yml, add new registry roles in the start of it and use the new site.yml in Makefile. Cleanup the leftovers.


annashipman:
Which story?

mtekel:
Updated.

keymon:
[Same comment as what we discussed adding postgresql](https://github.com/alphagov/tsuru-ansible/pull/127#commitcomment-12250020), here `aws_registry_vars.yml` and `gce_registry_vars.yml` are actually group variables. 

As we add more and more of these files the project directory is getting messy.

Appart of that, the naming of the files is not consistent:
- `platform-aws.yml` and `platform-gce.yml`: IaaS at the end
- `postgres_common_vars.yml` `postgres_master_vars.yml`: "role" at the beginning, specific context (master/common) later.
- But `gce_registry_vars.yml` and `aws_registry_vars.yml`: IaaS at the beggining (specific part) and "role" after.

Makes more sense  `registry_gce_vars.yml` and `registry_aws_vars.yml`

mtekel:
I wanted to make differences more explicit, hence I put the IAAS name in the beginning of the vars file (as IAAS is the main difference driver between the two vars files). And yes, this should be properly refactored into groups. This is the 1st step.

keymon:
I am going to rebase this from master in order to be able to review it without destroying all the environments, as I built them with https://github.com/alphagov/tsuru-ansible/pull/128 in place.

keymon:
About the PR description itself, I would explain better the context, motivation, related card and priorities.

keymon:
Tested after rebasing with master. Tested recreating the registry server, and works fine.

Once the comments above get answered I will be happy to merge it.

keymon:
I amend the commits 1c65a50960563bec1396562afb59aee00b6c8fd4 and d7e5aa70fa921379da6a6af587a8bd9d71db0653 to reflect the real actions

mtekel:
Regarding postgres globals and common: The variable split was done already before I pulled the vars from the playbook into common. Someone for some reason has decided to split the variables this way and I had no reason to change this. 

Regarding file naming: We should probably discuss the ansible approach we want to take (groups/files/globals etc.), then come up with some naming guidelines and stick with them. Currently it's ad-hoc based on what seems a better name.

keymon:
I disagree about:
- naming convention of `registry_gce_vars.yml / registry_aws_vars.yml` vs `gce_registry_vars.yml / aws_registry_vars.yml`. I do prefer the later.
- moving the variables to `globals.yml` as bfb0f288d84762afbd2a59c7526df0f9d6cdd107  . These variables are specific to the playbook of registry. I consider there should be either in:
  - in the `registry.yml` playbook, in a vars section as done in a lot of other playbooks.
  - in a separated file `registry_common_vars.yml` as in the postgres playbook.

keymon:
After last changes, I'm happy now :) :+1:

PR #130, opened 2015-07-20
[#99121602] Configure elasticsearch

mtekel:
Story: [#99121602](https://www.pivotaltracker.com/n/projects/1275640/stories/99121602)

Configure elasticsearch server, deploy broker (runs on the same server) and register elasticsearch service in tsuru.

Depends on https://github.com/alphagov/tsuru-terraform/pull/86
### How to test
1. Run terraform from [this](https://github.com/alphagov/tsuru-terraform/pull/86) PR against your environment - this adds elasticsearch server
2. Run ansible from this PR - that configures the server, adds broker and registers service

To verify service was correctly added:
1. Create service in tsuru `tsuru service-add elasticsearch mysearch -t myteam`
2. Bind the service to your app `tsuru service-bind mysearch -a myapp`
3. Check environment variables of your app, there should be `ELASTICSEARCH_HOST` and `ELASTICSEARCH_PORT` private variables present.

To test the service itself, you can try deploying the [digital marketplace search api](https://github.com/RichardKnop/digitalmarketplace-search-api.git). After deploying the app, you can check https://<your_digitalmarketplace_search_api_app>/_status which should tell if all is fine and if any digitalmarketplace-search-api related data are present in elasticsearch.
### Who can test:

Anyone but @mtekel and @jimconner


elasticsearch.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Install ES 1.6.0
+      apt:
+        deb: /tmp/elasticsearch-1.6.0.deb
+    - name: Start the service
+      service:
+        name: elasticsearch
+        state: started
+    - name: Git pull broker
+      git:
+        repo: https://github.com/RichardKnop/tsuru-elasticsearch-api.git
```
saliceti, line 30: Could we move it to an alphagov repository?

elasticsearch.yml
```diff
@@ -0,0 +1,41 @@
...
+        dest: /home/ubuntu/tsuru-elasticsearch-api
+      notify:
+        - Run tsuru-elasticsearch-api
+    - name: Install Python requirements
+      command: pip install -r //home/ubuntu/tsuru-elasticsearch-api/requirements.txt
```
saliceti: Remove double slash

saliceti:
These terraform and ansible changes are only for temporary performance tests, they are not meant for production. Do we really want these merged into our official platform?
I feel they should be made optional.

This can go as is for the time being. We will make our configuration more modular in #99509290.

saliceti:
Tested successfully with DM search application.

elasticsearch.yml
annashipman, line 30: We've had a conversation offline about this and think actually it would be better to build on the out of date [Elasticsearch API created by Globo](https://github.com/globocom/elastic-search-api) rather than use the one created for the spike. That one is perfectly fine for a spike, but if we are going to use it it would need a better README, clearer commit messages and ideally, tests, and building on the work that Globo have already done around this would put us in a better position.

mtekel:
Waiting on elasticsearch broker improvement. Will be re-opened when all changes are done in all other related repositories.

PR #131, opened 2015-07-21
[#95512034] Faster docker node deployment

mtekel:
Do the main and slow docker server installation in parallel on all servers. Then run the concurrent-ssh-bug affected actions in serial, node-by-node. This speeds up deployment of the platform considerably.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -36,11 +36,13 @@
       service: name=gandalf-server state=restarted
 
 - hosts: "{{ hosts_prefix }}-tsuru-docker*"
-  serial: 1
   sudo: yes
   roles:
     - docker_server
 
+- hosts: "{{ hosts_prefix }}-tsuru-docker*"
+  serial: 1
+  sudo: yes
   post_tasks:
     - name: Register node.
       shell: >
--- a/site.yml
+++ b/site.yml
@@ -5,11 +5,6 @@
   sudo: yes
   roles:
     - bennojoy.redis
-
-- hosts: "{{ hosts_prefix }}-tsuru-db"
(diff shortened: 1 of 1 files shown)


common.yml
```diff
@@ -3,11 +3,6 @@
   sudo: yes
   roles:
     - bennojoy.redis
-
-- hosts: "{{ hosts_prefix }}-tsuru-db"
-  sudo: yes
-  sudo_user: root
-  roles:
```
dcarley: The commit message doesn't mention why this changes. Which probably also means that it should be in a separate commit.

common.yml
```diff
@@ -38,12 +33,16 @@
     - name: restart gandalf
       service: name=gandalf-server state=restarted
 
+# Configure all docker nodes in parallel
```
dcarley: We should omit this comment because it's describing Ansible's default behaviour.

common.yml
```diff
@@ -38,12 +33,16 @@
...
 - hosts: "{{ hosts_prefix }}-tsuru-docker*"
-  serial: 1
   sudo: yes
   roles:
     - docker_server
 
+# Run this serially due to concurrent ssh issue
```
dcarley: If we're going to comment this then it should be the line above `serial: 1` to make it clear what the difference is and make it more likely that we remove the comment at the same time as the line.

dcarley:
The approach LGTM.

mtekel:
Updated the commits.

dcarley:
Will test after lunch

dcarley:
:+1:

PR #133, opened 2015-07-22
[#98640854] Add logic to handle locked app in test after hook

keymon:
Reference: [Docker nodes in demo environment uncontactable](https://www.pivotaltracker.com/n/projects/1275640/stories/98640854)
### Context

We found out that sometimes the smoke tests don't clean properly the apps, leaving them running in the platform. 

After investigating, we found that when running the test the application can be potentially locked in the
tsuru API server, for instance because the git push timeout or was disconnected. When this happens, the `after` hook will fail cleaning up the application. 

This code adds logic which will wait 5 seconds and retry if it fails deleting it because it is locked, giving a change to finish the running task. After 5 retries, it will force an unlock to delete it.
### How to review this PR?

You should be able to run the tests (check README.md):

```
make test-gce DEPLOY_ENV=<yourenv>
make test-aws DEPLOY_ENV=<yourenv>
```

and no applications are left behind:

```
$ tsuru app-list | grep sampleapp
$
```

If you want to trigger the retry when lock logic, you can kill the git command doing the push with: `killall git`
### Who can review this?

Anybody but @keymon or @jimconner

### DIFF
--- a/spec/command_line_helper.rb
+++ b/spec/command_line_helper.rb
@@ -12,7 +12,15 @@ class CommandLineHelper
   def wait()
     @tout.join if @tout
     @terr.join if @terr
-    @exit_status = @wait_thread.value.to_i >> 8 if @wait_thread
+    if @wait_thread
+      # Return code can be the the signal number if killed with signal
+      # or exit value shifted 8 bits if exited normally
+      if @wait_thread.value.signaled?
+        @exit_status = @wait_thread.value.to_i + 128
+      else
+        @exit_status = @wait_thread.value.to_i >> 8
+      end
+    end
     self
   end
 
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -60,7 +60,7 @@ describe "TsuruEndToEnd" do
       @workspace.tsuru_command.service_unbind(@sampleapp_db_instance, @sampleapp_name)
       @workspace.tsuru_command.service_remove(@sampleapp_db_instance)
       # Remove the application. Wait for the unlock if needed
-      retries=5
+      retries=25
       begin
         @workspace.tsuru_command.app_remove(@sampleapp_name)
         expect(@workspace.tsuru_command.stderr).to_not match /App locked by/
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -55,10 +55,21 @@ describe "TsuruEndToEnd" do
     end
 
     after(:all) do
+      # Remove previous state if needed.
       @workspace.tsuru_command.key_remove('rspec') # Remove previous state if needed
       @workspace.tsuru_command.service_unbind(@sampleapp_db_instance, @sampleapp_name)
-      @workspace.tsuru_command.service_remove(@sampleapp_db_instance) # Remove previous state if needed
-      @workspace.tsuru_command.app_remove(@sampleapp_name) # Remove previous state if needed
+      @workspace.tsuru_command.service_remove(@sampleapp_db_instance)
+      # Remove the application. Wait for the unlock if needed
+      retries=5
+      begin
+        @workspace.tsuru_command.app_remove(@sampleapp_name)
+        expect(@workspace.tsuru_command.stderr).to_not match /App locked by/
+      rescue RSpec::Expectations::ExpectationNotMetError
+        sleep 5
+        retry if (retries -= 1) > 0
+        @workspace.tsuru_command.app_unlock(@sampleapp_name)
+        @workspace.tsuru_command.app_remove(@sampleapp_name)
+      end
       @workspace.clean
     end
 
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -70,10 +70,10 @@ describe "TsuruEndToEnd" do
       end
       retries=25
       begin
+        sleep 1
         @workspace.tsuru_command.service_remove(@sampleapp_db_instance)
         expect(@workspace.tsuru_command.stderr).to_not match /This service instance is bound to at least one app/
       rescue RSpec::Expectations::ExpectationNotMetError
-        sleep 1
         retry if (retries -= 1) > 0
       end
       @workspace.clean
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -56,10 +56,8 @@ describe "TsuruEndToEnd" do
 
     after(:all) do
       # Remove previous state if needed.
-      @workspace.tsuru_command.key_remove('rspec') # Remove previous state if needed
-      @workspace.tsuru_command.service_unbind(@sampleapp_db_instance, @sampleapp_name)
-      @workspace.tsuru_command.service_remove(@sampleapp_db_instance)
-      # Remove the application. Wait for the unlock if needed
+      @workspace.tsuru_command.key_remove('rspec')
+      # Remove the app. Wait for the unlock if needed
       retries=25
       begin
         @workspace.tsuru_command.app_remove(@sampleapp_name)
@@ -70,6 +68,7 @@ describe "TsuruEndToEnd" do
         @workspace.tsuru_command.app_unlock(@sampleapp_name)
         @workspace.tsuru_command.app_remove(@sampleapp_name)
       end
+      @workspace.tsuru_command.service_remove(@sampleapp_db_instance)
       @workspace.clean
     end
 
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -115,20 +115,20 @@ describe "TsuruEndToEnd" do
       expect(@workspace.tsuru_command.stdout).to match /Instance .* is now bound to the app .*/
     end
 
-    it "Should be able to push the application" do
+    it "should be able to push the application" do
       git_url = @workspace.tsuru_command.get_app_repository(@sampleapp_name)
       expect(git_url).not_to be_nil
       @git_command.push(git_url)
       expect(@git_command.exit_status).to eql 0
     end
 
-    it "Should be able to connect to the applitation via HTTPS" do
+    it "should be able to connect to the applitation via HTTPS" do
       sampleapp_address = @workspace.tsuru_command.get_app_address(@sampleapp_name)
       response = URI.parse("https://#{sampleapp_address}/").open({ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE})
       expect(response.status).to eq(["200", "OK"])
     end
 
-    it "Should get log output in 3 seconds" do
+    it "should get log output in 3 seconds" do
       sampleapp_address = @workspace.tsuru_command.get_app_address(@sampleapp_name)
       query = "my_special_query_" + Time.now.to_i.to_s
       @workspace.tsuru_command.tail_app_logs(@sampleapp_name)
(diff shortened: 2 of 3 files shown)


mtekel:
Shouldn't total timeout (currently 5*5 seconds) be longer than default tsuru action timeout (which I think is 2 minutes, if not then perhaps 30s)? When you unlock app and remove it, some actions don't always happen. E.g. you could end up with app being removed, but app repo still existing. When you then try to add app with the same name, you can't, because repository already exists on gandalf. This way you are cleaning up the apps, but you are still leaking some objects elsewhere...

keymon:
About the timeout: 25 seconds should should be enough to deploy the sample application we use to deploy it. We don't want to make the test longer than this.

It is true that there are objects that might be leak in other points. But this change is just _one step in the right direction(tm)_, does not try to short out all issues but only solve one which caused some trouble: Increase the chances of delete an application if it is locked.

mtekel:
I think the app leaking doesn't happen all the time, so 2 minutes timeout would slow down only those tests that had issue. When you fix only this problem, then you remove the 1st issue (memory consumption by orphaned apps) that manifests itself early. It's then just a question of time when the next things starts failing. But it would be interesting to see what it will be and how exactly would that affect the platform...

keymon:
That is the point of the story [#98640854 Docker nodes in demo environment uncontactable](https://www.pivotaltracker.com/n/projects/1275640/stories/98640854)

The point of this PR is: Wait a while to give the app an opportunity to finish whatever is doing (why the lock is set), and if does not finish, remove the lock and remove the app.  If we don't add this PR, the test will finish and leave apps running.

Is your suggestion add 24 retries to sum a 2 minutes timeout? Can you give references of why 2 minutes is a good value?

Note that if we do so, we should tune the frequency we run the tests to at least 4 minutes [here](https://github.com/alphagov/multicloud-deploy/blob/master/site.yml#L147), to avoid jenkins adding too many jobs in the queue. Or we can just leave it as 2m frequency, and change it if it is needed.

mtekel:
2 mins come from here: https://github.com/tsuru/tsuru/blob/master/provision/docker/healthcheck.go#L67

If there's anything gone wrong with the app, tsuru will wait up to 2 mins before it starts cleanup/rollback. I am not sure what happens if you unlock and remove app before this ends. Might leave orphaned container without listing the app in the list of apps.

As far as jenkins goes, I don't think it would queue more than 1 job...

I don't think the failures are happening frequently, so this shouldn't be a big issue. I'd prefer proper cleanup with delay compared to just attempted cleanup with shorter delay. And if the issues solve themselves say after 10 or 20 seconds, then the longer max retry period would not impact anything. I would say if the issue doesn't resolve itself in 25 seconds, then it's something more complicated and it's worth waiting over 2 mins and attempt to clean it properly...

keymon:
OK, I will increase it to sum 120 seconds.

keymon:
> Might leave orphaned container without listing the app in the list of apps.

Actually this won't happen given [how tsuru works while deleting an application](https://github.com/tsuru/tsuru/blob/0.11.2/app/app.go#L308)

It is true that state management and locking in tsuru is quite complex and can drive to different situations and potential inconsistency, but that is not the point of this PR.

keymon:
Increased to 25 retries with 5 second delay.

dcarley:
@keymon I think this should be tagged to story [#99026674](https://www.pivotaltracker.com/story/show/99026674) rather than [#98640854](https://www.pivotaltracker.com/story/show/98640854). Also it's missing a hash in front of the number so the hook won't work :smile:

mtekel:
Did some testing, I was simulating git push failure by locally killing git client while it's doing push. The tests correctly detected this, but they exited immediately to commandline. There was no timeout whatsoever. It seems like the `after(:all)` part was never called:

```
TsuruEndToEnd
  healthchecks
    should return each component as WORKING
    should return status code of 200 to indicate all components are healthy
  deploying an application
    should not be able to login via HTTP
    should be able to login via HTTPS
    should be able to add the ssh key
    should be able to create an application
    should be able to create a service
    should be able to bind a service to an app
    Should be able to push the application (FAILED - 1)

Failures:

  1) TsuruEndToEnd deploying an application Should be able to push the application
     Failure/Error: @git_command.push(git_url)
     Errno::ENOENT:
       No such file or directory - chdir
     # ./spec/command_line_helper.rb:32:in `execute_helper_async'
     # ./spec/command_line_helper.rb:50:in `execute_helper'
     # ./spec/git_helper.rb:65:in `push'
     # ./spec/endtoend/endtoend_spec.rb:115:in `block (3 levels) in <top (required)>'

Finished in 9.21 seconds (files took 0.10511 seconds to load)
9 examples, 1 failure

Failed examples:

rspec ./spec/endtoend/endtoend_spec.rb:112 # TsuruEndToEnd deploying an application Should be able to push the application
```

dcarley:
> It seems like the after(:all) part was never called:

Not sure if I can tell that from the output. Is there an app left in the output of `tsuru app-list`? Would it have locked the app if it didn't complete the `git push` (and thus didn't create a new image or any new units)?

keymon:
It is better you call it with: `make test-aws DEPLOY_ENV=... VERBOSE=true`

That will show if the  `after(:all)` is called

keymon:
Also, I think you killed the `git clone`, not the `git push`. There is a bug in this smoke test that if the clone fails, it carries on running and it fails with the error you got when trying to do a `git push` (`Errno::ENOENT: No such file or directory - chdir`)

keymon:
@mtekel Sorry, I did not know about  [#99026674](https://www.pivotaltracker.com/story/show/99026674). I started this change motivated by [#98640854](https://www.pivotaltracker.com/story/show/98640854)

mtekel:
So in the 1st case I have killed git way too early, likely even before push could start. Tried now with verbose:

```
stderr: remote: Preparing to unpack .../libgssrpc4_1.12+dfsg-2ubuntu5.1_amd64.deb ...
stderr: remote: Unpacking libgssrpc4:amd64 (1.12+dfsg-2ubuntu5.1) ...
stderr: remote: Selecting previously unselected package libkadm5clnt-mit9:amd64.
stderr: remote: Preparing to unpack .../libkadm5clnt-mit9_1.12+dfsg-2ubuntu5.1_amd64.deb ...
stderr: remote: Unpacking libkadm5clnt-mit9:amd64 (1.12+dfsg-2ubuntu5.1) ...
Exit code: 0
    Should be able to push the application
Executing: HOME='/tmp/tsuru-command20150723-51425-10ikc34' tsuru app-info -a sampleapp1437646781
stdout: Application: sampleapp1437646781
stdout: Repository: git@russ-gandalf.tsuru.paas.alphagov.co.uk:sampleapp1437646781.git
stdout: Platform: python
stdout: Teams: admin
stdout: Address: sampleapp1437646781.russ-hipache.tsuru.paas.alphagov.co.uk
stdout: Owner: administrator@gds.tsuru.gov
stdout: Team owner: admin
stdout: Deploys: 0
stdout: Pool: default
stdout: Units: 1
stdout: +------------+----------+---------------+------+----+
stdout: | Unit       | State    | Host          | Port | IP |
stdout: +------------+----------+---------------+------+----+
stdout: | d0e14d6f20 | building | 10.128.11.233 |      |    |
stdout: +------------+----------+---------------+------+----+
stdout: 
stdout: Service instances: 1
stdout: +------------+---------------+
stdout: | Service    | Instance      |
stdout: +------------+---------------+
stdout: | postgresql | tst2876467341 |
stdout: +------------+---------------+
stdout: 
stdout: App Plan:
stdout: +---------------+--------+------+-----------+--------+---------+
stdout: | Name          | Memory | Swap | Cpu Share | Router | Default |
stdout: +---------------+--------+------+-----------+--------+---------+
stdout: | autogenerated | 0 MB   | 0 MB | 100       |        | false   |
stdout: +---------------+--------+------+-----------+--------+---------+
stdout: 
Exit code: 0
    Should be able to connect to the applitation via HTTPS (FAILED - 1)
Executing: HOME='/tmp/tsuru-command20150723-51425-10ikc34' tsuru key-remove rspec -y
stdout: Key "rspec" successfully removed!
Exit code: 0
Executing: HOME='/tmp/tsuru-command20150723-51425-10ikc34' tsuru service-unbind tst2876467341 -a sampleapp1437646781
stderr: Error: App locked by tsr, running POST /apps/sampleapp1437646781/repository/clone. Acquired in 2015-07-23T10:20:06Z
Exit code: 1
Executing: HOME='/tmp/tsuru-command20150723-51425-10ikc34' tsuru service-remove tst2876467341 -y
stderr: Error: This service instance is bound to at least one app. Unbind them before removing it
Exit code: 1
Executing: HOME='/tmp/tsuru-command20150723-51425-10ikc34' tsuru app-remove -a sampleapp1437646781 -y
stderr: Error: App locked by tsr, running POST /apps/sampleapp1437646781/repository/clone. Acquired in 2015-07-23T10:20:06Z
Exit code: 1
Executing: HOME='/tmp/tsuru-command20150723-51425-10ikc34' tsuru app-remove -a sampleapp1437646781 -y
stdout: App "sampleapp1437646781" successfully removed!
Exit code: 0

Failures:

  1) TsuruEndToEnd deploying an application Should be able to connect to the applitation via HTTPS
     Failure/Error: response = URI.parse("https://#{sampleapp_address}/").open({ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE})
     OpenURI::HTTPError:
       502 Bad Gateway
     # ./spec/endtoend/endtoend_spec.rb:121:in `block (3 levels) in <top (required)>'

Finished in 44.42 seconds (files took 0.10191 seconds to load)
10 examples, 1 failure

Failed examples:

rspec ./spec/endtoend/endtoend_spec.rb:119 # TsuruEndToEnd deploying an application Should be able to connect to the applitation via HTTPS
```

Cleanup happens correctly.

keymon:
Actually, @mtekel, you found a different issue :) The service is not cleaned up:

```
Executing: HOME='/tmp/tsuru-command20150723-51425-10ikc34' tsuru service-unbind tst2876467341 -a sampleapp1437646781
stderr: Error: App locked by tsr, running POST /apps/sampleapp1437646781/repository/clone. Acquired in 2015-07-23T10:20:06Z
Exit code: 1
```

I shall add logic around that.

mtekel:
Findings:
- even though I kill git mid deploy, which causes tsuru client to stop immediately, it returns 0 - which means OK. The deploy test doesn't fail. In reality, deploy of the app continues in the background.
- services are still leaking, should be inside retry loop

This opens a question: What if the app takes 15 minutes do deploy... and we unlock and remove it after 2 mins? Will that leave some orphaned container anywhere? Will that leave git repos that still exist and prevent app of the same name to be created again? Will that leave some data in the tsuru database (mongo)? Etc...

Still, this is a good improvement. Adding all created resources cleanup into retry loop would solve a most of leaks we currently have.

keymon:
> What if the app takes 15 minutes do deploy... and we unlock and remove it after 2 mins? Will that leave some orphaned container anywhere? Will that leave git repos that still exist and prevent app of the same name to be created again? Will that leave some data in the tsuru database (mongo)? Etc...

Good questions! We should investigate it further, and you can find  some related comments in [#98652142](https://www.pivotaltracker.com/n/projects/1275640/stories/98652142)). 

Anyway, that is out of the scope of this PR, we can just assume/hope that it will finish before 2m :(

> Adding all created resources cleanup into retry loop would solve a lot of leaks we currently have.

Will do.

keymon:
I fixed the issue of getting a exit code == 0 when killing the child command.

mtekel:
Killing git during the deploy now causes the push app test to correctly fail immediately.

The last issue: service unbind doesn't happen immediately after app-remove, so the subsequent service-remove fails. Putting service remove in the retry loop should fix this.

mtekel:
OK, this all works now nicely, merging.

PR #134, opened 2015-07-22
[#98824506] Restore postgres to a different IP

saliceti:
We may need to restore a postgres database to a different host with a different IP address. This means we will need to repoint the postgres server instance to the new IP address and ensure that any apps consuming the database are still working correctly.

Since we already have a hot standby host, we test by promoting the standby to master.

**How to review**
- Build a new environment or update one with the latest terraform so it includes the Postgres Master DNS record
- In the case of a previous environment, remove the postgresapi application
- Run ansible from this PR
- Deploy the [flask-postgres application](https://github.com/alphagov/flask-sqlalchemy-postgres-heroku-example) (with postgres service)
- Enter a few posts
- Stop the master postgres node (postgres-0)
- Manually change the DNS record _env_-postgres-master.tsuru[2].paas.alphagov.co.uk to point to the standby node (postgres-1)
- Wait for the flask app to see the new IP (1-2 min). Ex:

``` bash
tsuru app-run "ping smashmouth-postgres-master.tsuru.paas.alphagov.co.uk" -a flask
```
- Reload the flask app. You should be able to read the previous content but you can't write a new post
- Run ansible on Postgres current standby node. Ex:

``` bash
make aws DEPLOY_ENV=smashmouth ARGS='-l tag_Name_smashmouth-tsuru-postgres-1'
```
- Reload the flask app. You should be able to read the previous content and write new posts

**Who should review**
Anyone but @actionjack and @saliceti

### DIFF
--- a/platform-aws.yml
+++ b/platform-aws.yml
@@ -21,7 +21,7 @@ mongodb_host_name: "{{ hosts_prefix }}-tsuru-db"
 mongodb_host: "{{ hostvars[groups[mongodb_host_name][0]][ip_field_name] }}"
 
 postgres_host_name: "{{ hosts_prefix }}-tsuru-postgres-0"
-postgres_host: "{{ hostvars[groups[postgres_host_name][0]][ip_field_name] }}"
+postgres_host: "{{ deploy_env }}-postgres-master.{{ domain_name }}"
 postgres_master: "{{ hostvars[groups[postgres_host_name][0]][ip_field_name] }}"
 
 influxdb_host_name: "{{ hosts_prefix }}-influx-grafana"
--- a/platform-gce.yml
+++ b/platform-gce.yml
@@ -21,7 +21,7 @@ mongodb_host_name: "{{ hosts_prefix }}-tsuru-db"
 mongodb_host: "{{ hostvars[mongodb_host_name][ip_field_name] }}"
 
 postgres_host_name: "{{ hosts_prefix }}-tsuru-postgres-0"
-postgres_host: "{{ hostvars[postgres_host_name][ip_field_name] }}"
+postgres_host: "{{ deploy_env }}-postgres-master.{{ domain_name }}"
 postgres_master: "{{ hostvars[postgres_host_name][ip_field_name] }}"
 
 influxdb_host_name: "{{ hosts_prefix }}-influx-grafana"


postgres.yml
```diff
@@ -6,8 +6,17 @@
...
+      shell: "dig +short {{ postgres_host }}"
+      register: postgres_master_IP_dig
+    - name: "PostgreSQL | Set Postgres Master IP fact"
+      set_fact:
+        postgres_master: "{{ postgres_master_IP_dig.stdout }}"
     - include_vars: postgres_master_vars.yml
       when: "ansible_default_ipv4.address == postgres_master"
+    - name: PostgreSQL Master | Delete recovery.conf
+      file: path=/var/lib/postgresql/9.3/main/recovery.conf state=absent
+      when: "ansible_default_ipv4.address == postgres_master"
```
mtekel, line 20: @russss  mentioned that you should not need to remove recovery. When switching to master it would be automatically removed by postgres.

group_vars/all/globals.yml
```diff
@@ -34,3 +34,5 @@ wal_e_pgdata_dir: "/var/lib/postgresql/{{ postgresql_version }}/main/"
 postgresql_archive_command: >
   /usr/bin/envdir {{ wal_e_envdir }} /usr/local/bin/wal-e
   {% if wal_e_aws_instance_profile %}--aws-instance-profile{% endif %} wal-push %p
+
+postgres_host: "{{ deploy_env }}-postgres-master.{{ domain_name }}"
```
mtekel, line 38: I'd call this postgres_master and the other variable postgres_master_ip

postgres.yml
saliceti, line 20: I tested first without removing recovery.conf and it was readonly. Maybe Russ meant keeping the file but setting hot_standby variable to off.
I'm in favour of removing the file altogether.

group_vars/all/globals.yml
keymon, line 38: Yes, I agree. These names might be confusing. 

Currently:
- `postgres_host`: DNS alias of the postgres master node
- `postgres_master`: IP of the postgres master node.

I suggest something like:
- `postgres_master_host`: DNS alias of the postgres master node
- `postgres_master_ip`: IP of the postgres master node.

russss:
The correct method of forcing a failover to a standby node is to trigger it using `pg_ctl promote` or touching a `trigger_file` referenced in the `recovery.conf`. Simply deleting the `recovery.conf` isn't recommended.

The other issue is that, once a standby is promoted to a master, a new WAL-E base backup will need to be taken (as the timeline ID has changed).

saliceti:
Closing for now to improve the failover process according to @russss

PR #135, opened 2015-07-22
[#99121602] Add elasticsearch service

mtekel:
Configure elasticsearch service - install server, deploy broker to tsuru and register the service.  Elasticsearch service is required for the [performance tests](https://www.pivotaltracker.com/n/projects/1275640/stories/99395802). 

This PR depends on https://github.com/alphagov/tsuru-terraform/pull/91.

### DIFF
--- a/elasticsearch.yml
+++ b/elasticsearch.yml
@@ -1,0 +1,19 @@
+---
+
+- hosts: "{{ hosts_prefix }}-tsuru-elasticsearch"
+  sudo: yes
+  sudo_user: root
+  tasks:
+    - name: Elasticsearch | Install java
+      apt: name=openjdk-7-jre-headless state=present update_cache=yes
+    - name: Elasticsearch | Download ES 1.6.0
+      get_url:
+        url: https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.6.0.deb
+        dest: /tmp/elasticsearch-1.6.0.deb
+    - name: Elasticsearch | Install ES 1.6.0
+      apt:
+        deb: /tmp/elasticsearch-1.6.0.deb
+    - name: Elasticsearch | Start the service
+      service:
+        name: elasticsearch
+        state: started
--- a/group_vars/all/globals.yml
+++ b/group_vars/all/globals.yml
@@ -36,3 +36,6 @@ postgresql_archive_command: >
   {% if wal_e_aws_instance_profile %}--aws-instance-profile{% endif %} wal-push %p
 
 postgres_master_host: "{{ deploy_env }}-postgres-master.{{ domain_name }}"
+
+elasticsearch_host_name: "{{ hosts_prefix }}-tsuru-elasticsearch"
+elasticsearch_url: "{{ deploy_env }}-elasticsearchapi.{{ domain_name }}"
--- a/group_vars/all/globals.yml
+++ b/group_vars/all/globals.yml
@@ -38,4 +38,4 @@ postgresql_archive_command: >
 postgres_master_host: "{{ deploy_env }}-postgres-master.{{ domain_name }}"
 
 elasticsearch_host_name: "{{ hosts_prefix }}-tsuru-elasticsearch"
-elasticsearch_url: "{{ deploy_env }}-elasticsearchapi.{{ domain_name }}"
+elasticsearch_api: "{{ deploy_env }}-elasticsearchapi.{{ domain_name }}"
--- a/platform-aws.yml
+++ b/platform-aws.yml
@@ -20,6 +20,8 @@ gandalf_host_internal: "{{ hostvars[groups[gandalf_host_name][0]][ip_field_name]
 mongodb_host_name: "{{ hosts_prefix }}-tsuru-db"
 mongodb_host: "{{ hostvars[groups[mongodb_host_name][0]][ip_field_name] }}"
 
+elasticsearch_host: "{{ hostvars[groups[elasticsearch_host_name][0]][ip_field_name] }}"
+
 influxdb_host_name: "{{ hosts_prefix }}-influx-grafana"
 influxdb_url: "http://{{ hostvars[groups[influxdb_host_name][0]][ip_field_name] }}:8086"
 
--- a/post-install.yml
+++ b/post-install.yml
@@ -178,6 +178,66 @@
         crane create ~/postgresapi.yml
       when: "not 'postgresql' in service_list.stdout"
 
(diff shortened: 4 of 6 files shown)


post-install.yml
```diff
@@ -169,15 +171,70 @@
...
-    - name: Register service
+    - name: Post Install Tasks | Register service
       shell: >
         crane create ~/postgresapi.yml
       when: "not 'postgresql' in service_list.stdout"
 
+    # Elasticsearch
+
+    - name: Post Install Tasks | Tsuru app-create elasticsearchapi
+      ignore_errors: yes
```
saliceti: Why ignore_errors? We don't want to continue if there is a failure.
I don't understand why we have more _ignore_errors: yes_ above

post-install.yml
```diff
@@ -169,15 +171,70 @@
...
+      shell: >
+        tsuru app-create elasticsearchapi python -t admin
+      when: "not 'elasticsearchapi' in tsuru_apps.stdout"
+
+    - name: Post Install Tasks | Git pull elasticsearchapi
+      git:
```
saliceti: Can we pin the version with refspec

saliceti:
Error running ansible

```
TASK: [Elasticsearch | Start the service] ************************************* 
failed: [10.128.11.139] => {"failed": true}
msg: Could not find any executable java binary. Please install java in your PATH or set JAVA_HOME
```

post-install.yml
```diff
@@ -169,15 +171,70 @@
...
+
+    - name: Post Install Tasks | Tsuru app-info elasticsearchapi
+      shell: >
+        tsuru app-info -a elasticsearchapi
+      register: elasticsearchapi_app_info
+
+    - name: Post Install Tasks | Tsuru set environment variables for elasticsearchapi
+      shell: >
+        tsuru env-set -a elasticsearchapi ELASTICSEARCH_HOST={{ elasticsearch_host }};
+      when: "not 'Everything up-to-date' in elasticsearchapi_push_result.stderr"
```
saliceti: - Move this before the git push to avoid restarting the application
- Change the _when_ to check with _env-get_

post-install.yml
```diff
@@ -169,15 +171,70 @@
...
+
+    - name: Post Install Tasks | Add <deploy_env>-elasticsearchapi.<domain_name> cname to elasticsearchapi app
+      shell: >
+        tsuru cname-add {{ deploy_env }}-elasticsearchapi.{{ domain_name }} -a elasticsearchapi
+      when: "not '-elasticsearchapi.' in elasticsearchapi_app_info.stdout"
```
saliceti: Weak test. I would use the full CNAME

post-install.yml
```diff
@@ -169,15 +171,70 @@
...
+
+    - name: Post Install Tasks | Generate elasticsearchapi.yml
+      copy:
+        content: |
+          id: elasticsearch
+          username: admin
+          password: password
+          endpoint:
+            production: https://{{ deploy_env }}-elasticsearchapi.{{ domain_name }}
```
saliceti: Create an `elasticsearch_host` variable in globals.yaml for `{{ deploy_env }}-elasticsearchapi.{{ domain_name }}` to avoid repetition

elasticsearch.yml
```diff
@@ -0,0 +1,19 @@
+---
+
+- hosts: "{{ hosts_prefix }}-tsuru-elasticsearch"
+  sudo: yes
+  sudo_user: root
+  tasks:
+    - name: Elasticsearch | Install java
+      apt: name=openjdk-7-jre-headless state=present update_cache=yes
```
saliceti, line 8: Can you pin the package version?

group_vars/all/globals.yml
```diff
@@ -36,3 +36,6 @@ postgresql_archive_command: >
   {% if wal_e_aws_instance_profile %}--aws-instance-profile{% endif %} wal-push %p
 
 postgres_master_host: "{{ deploy_env }}-postgres-master.{{ domain_name }}"
+
+elasticsearch_host_name: "{{ hosts_prefix }}-tsuru-elasticsearch"
+elasticsearch_url: "{{ deploy_env }}-elasticsearchapi.{{ domain_name }}"
```
saliceti: This is not a URL, I would rename as elasticsearchapi_host

platform-aws.yml
```diff
@@ -20,6 +20,8 @@ gandalf_host_internal: "{{ hostvars[groups[gandalf_host_name][0]][ip_field_name]
 mongodb_host_name: "{{ hosts_prefix }}-tsuru-db"
 mongodb_host: "{{ hostvars[groups[mongodb_host_name][0]][ip_field_name] }}"
 
+elasticsearch_host: "{{ hostvars[groups[elasticsearch_host_name][0]][ip_field_name] }}"
```
saliceti, line 23: Rename as elasticsearch_ip

post-install.yml
```diff
@@ -169,15 +166,74 @@
...
+        git push {{ elasticsearchapi_git_remote.stdout }} master
+      args:
+        chdir: /tmp/elasticsearchapi
+
+    - name: Post Install Tasks | Tsuru app-info elasticsearchapi
+      shell: >
+        tsuru app-info -a elasticsearchapi
+      register: elasticsearchapi_app_info
+
+    - name: Post Install Tasks | Add <deploy_env>-elasticsearchapi.<domain_name> cname to elasticsearchapi app
```
saliceti, line 217: Use the variable here too instead of `<deploy_env>-elasticsearchapi.<domain_name>`

group_vars/all/globals.yml
mtekel: <server_type>_host is our usual nomenclature for that type of value, I would not like to change it just for elasticsearch.
This is also reason why I would not like to change _host to _ip... . It seems reasonable, but then I should rename all the variables in the proper way... (something _host_name -> something_server_name; something_host -> something_ip; and finally use something_host for host names).

saliceti: At least rename elasticsearch_url.
We can create a new story to standardise variable names.

post-install.yml
mtekel, line 217: I've tried that with {{ }} and {% %}, but it didn't work. They say they have just fixed that: https://github.com/ansible/ansible/issues/10347

So when we decide to do ansible upgrade then we can refactor this (and also e.g. postgres_api url).

saliceti:
Tested successfully with digital marketplace search API. Merging

PR #139, opened 2015-07-24
[#99133134] fix test that checks rebind of services

keymon:
## Context

The test `should be able to unbind and bind a service to an app` was pending [due a bug in tsuru postgres-api`](https://github.com/tsuru/postgres-api/issues/1). 

~~This bug [has been fixed](https://github.com/tsuru/postgres-api/pull/12) so we can remove it from pending.~~ From the comments from Michael, it seems the bug is not really fixed. But some commits in this PR are still valid.

It was necessary to fix some minor bugs in the test itself.
## How to review this?

Just run the tests as described in the README.md: `make test-aws DEPLOY_ENV=<env> VERBOSE=true` or `make test-gce DEPLOY_ENV=<env> VERBOSE=true`

~~The test `should be able to unbind and bind a service to an app` should be successful.~~

The test `should be able to unbind and bind a service to an app` should be pending, but fail with a `stderr: Error: Failed to bind the instance "tst3507577341" to the app "sampleapp1437757051": role "tst3507577234c09" already exists` error.
## Who should review this

Anyone but @keymon

### DIFF
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -144,15 +144,7 @@ describe "TsuruEndToEnd" do
       expect(@workspace.tsuru_command.stdout).to include query
     end
 
-    it "should be able to connect to the applitation via HTTPS with a valid cert" do
-      pending "We don't have a certificate for this :)"
-      sampleapp_address = @workspace.tsuru_command.get_app_address(@sampleapp_name)
-      response = URI.parse("https://#{sampleapp_address}/").open()
-      expect(response.status).to eq(["200", "OK"])
-    end
-
     it "should be able to unbind and bind a service to an app" do
-      pending "There is already a bug filed: https://github.com/tsuru/postgres-api/issues/1"
       @workspace.tsuru_command.service_unbind(@sampleapp_db_instance, @sampleapp_name)
       expect(@workspace.tsuru_command.exit_status).to eql 0
       retries=5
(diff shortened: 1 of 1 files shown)


mtekel:
There is issue with the postgresapi:
- when you bind app, it creates role
- this role is set as owner in database called `<service_instance_name>`, e.g. flaskdb or however you named your service instance
- when you unbind, removal of the role is attempted, but this is blocked due to that role being owner of the table schemas:

```
postgres=# drop role flaskdbe79681 ;
ERROR:  role "flaskdbe79681" cannot be dropped because some objects depend on it
DETAIL:  2 objects in database flaskdb
```
- if there are any more apps binding this service, they can unbind because their role is not the owner and is not blocked

This is postgresapi issue. And because of this, this new test will keep failing (until the issue is fixed).

mtekel:
To reproduce:

```
git clone git@github.com:alphagov/flask-sqlalchemy-postgres-heroku-example.git flask
cd flask
tsuru app-create flask python
tsuru service-add postgresql flaskdb
tsuru service-bind flask flaskdb
git push $(tsuru app-info -a flask | sed -n 's/Repository: //p')  
tsuru service-unbind -a flask flaskdb
tsuru service-bind -a flask flaskdb # fails

tsuru app-create flask2 python
tsuru service-bind -a flask2 flaskdb 
git push $(tsuru app-info -a flask2 | sed -n 's/Repository: //p')  
tsuru service-unbind -a flask2 flaskdb
tsuru service-bind -a flask2 flaskdb # works
```

RichardKnop:
:+1:

keymon:
One possible solution might be the one explained in http://stackoverflow.com/a/3023840. If you reassign the objects, you can drop the role:

```
flaskdb=# \d+
                             List of relations
Schema |     Name     |   Type   |     Owner     |    Size    | Description 
--------+--------------+----------+---------------+------------+-------------
public | entry        | table    | flaskdbe79681 | 0 bytes    | 
public | entry_id_seq | sequence | flaskdbe79681 | 8192 bytes | 
(2 rows)

flaskdb=# REASSIGN OWNED BY flaskdbe79681 TO flaskdb
;
REASSIGN OWNED
flaskdb=# \d+
                          List of relations
Schema |     Name     |   Type   |  Owner  |    Size    | Description 
--------+--------------+----------+---------+------------+-------------
public | entry        | table    | flaskdb | 0 bytes    | 
public | entry_id_seq | sequence | flaskdb | 8192 bytes | 
(2 rows)

```

We should report this bug back :(

mtekel:
Now this works, merging.

PR #140, opened 2015-07-24
[#99500102] Permission denied (publickey) on CI build

saliceti:
We assume this error happens when the local SSH keys are deleted or changed. The previous behaviour doesn't check if the remote key in Tsuru has the same content so we may end up out of sync.

Now we compare the content of local and remote keys so the new key will be uploaded if it is different.
There is no need to delete the remote key as Tsuru will replace the key of the same name if the content is different.
# How to review
- Run Ansible on a clean environment: it should create the keys and upload the public key
- Rerun Ansible: it should do nothing
- Delete the keys: it should create the keys and upload the public key
# Who can review

Anyone but @saliceti

### DIFF
--- a/post-install.yml
+++ b/post-install.yml
@@ -68,12 +68,15 @@
 
     - name: Post Install Tasks | Tsuru key-list
       shell: >
-        tsuru key-list
+        tsuru key-list -n
       register: tsuru_keys
+    - name: Post Install Tasks | Get public key content
+      shell: cat ~/.ssh/tsuru_deployer.pub
+      register: public_key_content
     - name: Post Install Tasks | Tsuru add ssh key
       shell: >
         tsuru key-add tsuru_deployer ~/.ssh/tsuru_deployer.pub
-      when: "not 'tsuru_deployer' in tsuru_keys.stdout"
+      when: "not public_key_content.stdout in tsuru_keys.stdout"
 
     - name: Post Install Tasks | Tsuru app-list
       shell: >
--- a/post-install.yml
+++ b/post-install.yml
@@ -73,6 +73,10 @@
     - name: Post Install Tasks | Get public key content
       shell: cat ~/.ssh/tsuru_deployer.pub
       register: public_key_content
+    - name: Post Install Tasks | Tsuru remove old key
+      shell: >
+        tsuru key-remove -y tsuru_deployer
+      when: "not public_key_content.stdout in tsuru_keys.stdout and 'tsuru_deployer' in tsuru_keys.stdout"
     - name: Post Install Tasks | Tsuru add ssh key
       shell: >
         tsuru key-add tsuru_deployer ~/.ssh/tsuru_deployer.pub
--- a/post-install.yml
+++ b/post-install.yml
@@ -1,4 +1,4 @@
-- hosts: "{{ hosts_prefix }}-tsuru-api*[0]"
+- hosts: "{{ hosts_prefix }}-tsuru-api-0"
   sudo: yes
   tasks:
     - name: Post Install Tasks | Install git client
--- a/post-install.yml
+++ b/post-install.yml
@@ -66,17 +66,17 @@
         tsuru-admin platform-add static -d https://raw.github.com/tsuru/basebuilder/master/static/Dockerfile
       when: "not 'static' in tsuru_platforms.stdout"
 
-    - name: Post Install Tasks | Tsuru key-list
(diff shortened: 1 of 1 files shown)


mtekel:
I do like the extended key matching test. However there is a catch with key update: tsuru allows you to overwrite key only if you set it to a new value. If you try to set it back to original value, it fails (it remembers what it was):

```
$ ssh-keygen -b 2048 -t rsa -f ~/.ssh/key_a -q -N ""
$ ssh-keygen -b 2048 -t rsa -f ~/.ssh/key_b -q -N ""
$ ssh-keygen -y -f ~/.ssh/key_a >~/.ssh/key_a.pub
$ ssh-keygen -y -f ~/.ssh/key_b >~/.ssh/key_b.pub

$ tsuru key-add mykey ~/.ssh/key_a.pub
Key "mykey" successfully added!
$ tsuru key-add mykey ~/.ssh/key_b.pub
Key "mykey" successfully added!
$ tsuru key-add mykey ~/.ssh/key_a.pub
Error: Key already exists.
```

Nevertheless, even though the old keys are not listed in the key-list output, they are still present on gandalf and do still work. I am not sure if this is intended. How can this happen in our case? We have 2 api servers, once post-install runs on the 1st one, then on 2nd one (updating the key) and then again on 1st one, failing to update the key. You can pin post-install to api-0 (as it is expected to run on this host), by changing `- hosts: "{{ hosts_prefix }}-tsuru-api*[0]"` to `- hosts: "{{ tsuru_api_host_name }}"`. See https://github.com/alphagov/tsuru-ansible/pull/137 for more details.

I also think you could just remove the key and add it again if the content's don't match. Running the tasks on api-0 doesn't guarantee that key doesn't get changed (could be e.g. removed during reboots).

Alternatively, you could simply add `ignore_errors: yes` to the key update task, with the comment why is it OK to ignore errors. If there are any real errors, they would likely prevent subsequent steps (app push) and the playbook would fail anyway.

mtekel:
Tsuru issue opened here: https://github.com/tsuru/tsuru/issues/1249

dcarley:
> You can pin post-install to api-0 (as it is expected to run on this host), by changing - hosts: "{{ hosts_prefix }}-tsuru-api*[0]" to - hosts: "{{ tsuru_api_host_name }}". See #137 for more details.

Also see a2ecb0426d6d22a8cd3658b6b8d43ff69308b556 in #141.

post-install.yml
```diff
@@ -68,12 +68,19 @@
...
-        tsuru key-list
+        tsuru key-list -n
       register: tsuru_keys
+    - name: Post Install Tasks | Get public key content
+      shell: cat ~/.ssh/tsuru_deployer.pub
+      register: public_key_content
+    - name: Post Install Tasks | Tsuru remove old key
+      shell: >
+        tsuru key-remove -y tsuru_deployer
+      when: "not public_key_content.stdout in tsuru_keys.stdout and 'tsuru_deployer' in tsuru_keys.stdout"
```
dcarley: This condition feels a bit loose. What do you think of removing the `and 'tsuru_deployer' in tsuru_keys.stdout` part and instead adding `| awk '$2 == "tsuru_deployer"'` to the `tsuru key-list -n` command, to pre-filter the results?

mtekel:
Updated...

post-install.yml
```diff
@@ -66,14 +66,22 @@
         tsuru-admin platform-add static -d https://raw.github.com/tsuru/basebuilder/master/static/Dockerfile
       when: "not 'static' in tsuru_platforms.stdout"
 
-    - name: Post Install Tasks | Tsuru key-list
+    - name: Post Install Tasks | Tsuru list tsuru_deployer key
       shell: >
-        tsuru key-list
+        tsuru key-list -n | awk '$2 == "tsuru_deployer" {print $5}'
```
dcarley: I think the `{print $5}` should be removed. It omits the key type (`ssh-rsa`) and causes the tasks to run every time:

```
TASK: [Post Install Tasks | Tsuru remove old key] *****************************
changed: [10.128.11.176]

TASK: [Post Install Tasks | Tsuru add ssh key] ********************************
changed: [10.128.11.176]
```

post-install.yml
```diff
@@ -66,14 +66,22 @@
...
       register: tsuru_keys
+    - debug: var=tsuru_keys
```
dcarley: Was this committed by accident?

dcarley:
:+1:

PR #141, opened 2015-07-27
[#99771782] De-dupe AWS/GCE variables

dcarley:
#### Add self-referencing group to gce.py

This is a hack so that we can de-duplicate the variables in
`platform-aws.yml` and `platform-gce.yml` by referring to the `hostvars` of
other nodes in a consistent way, e.g.

```
`hostvars[groups["{{ hosts_prefix }}-tsuru-api-0"][0]]`
```

This pattern is necessary on AWS because the `hostvars` dictionary is keyed
by the instance IP rather than it's `Name` tag and tags are not unique
across instances.

It might have been possible to do the same using jinja2 filters to select
entries with the appropriate attributes, but it looked to produce much more
complicated code.

I don't normally comment code, but this seems warranted since it's not at
all obvious why the code is doing that thing and we may want to find a
better solution in the future.
#### Use deploy_env instead of hosts_prefix for API LB

It just so happens that these two variables are the same for GCE. But for
simplicity and consistency with AWS, we should use `deploy_env`, which only
ever contains the name of the environment.
#### De-duplicate vars in platform-{aws,gce}.yml

The preceding commit allows us to refer to `hostvars` in the same way on AWS
and GCE. So we can move all of these variables to `group_vars` and not
duplicate them. Except for `tsuru_api_internal_lb` which is different.
#### Don't use var in hosts pattern for first API node

The preceding commit moved this variable from `platform-{aws,gce}.yml` to
`group_vars/all/globals.yml`. This seems to prevent Ansible from
interpolating this variable correctly:

```
PLAY [{{ tsuru_api_host_name }}] **********************************************
skipping: no hosts matched
```

Fix this by changing to a hosts pattern that only matches the first API
node. This is in keeping with the spirit of 3ff22e4, but unfortunately
leaves us with some duplication.

---

There was a suggestion of renaming the variables to `_ip` and `_server_name` in the story, but I think that's enough changes for a single PR. I'll either do that after or create a separate story.

### DIFF
--- a/gce.py
+++ b/gce.py
@@ -300,6 +300,9 @@ class GceInventory(object):
             if groups.has_key(stat): groups[stat].append(name)
             else: groups[stat] = [name]
 
+            # FIXME: Hack for compatibility with `ec2.py`
+            groups[name] = [name]
+
         groups["_meta"] = meta
 
         return groups
--- a/group_vars/all/globals.yml
+++ b/group_vars/all/globals.yml
@@ -1,6 +1,11 @@
 ---
 
+redis_host_name: "{{ hosts_prefix }}-tsuru-db"
+redis_host: "{{ hostvars[groups[redis_host_name][0]][ip_field_name] }}"
 redis_port: 6379
+
+tsuru_api_host_name: "{{ hosts_prefix }}-tsuru-api-0"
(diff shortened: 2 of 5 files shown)


mtekel:
Applied successfully against both existing and new builds in AWS and GCE. Merging...

PR #143, opened 2015-07-28
Smoke test to unbind and bind service enabled.

RichardKnop:
# What

I have enabled the smoke test that tests unbinding
and binding of a service. It works now since the issue
in Postgres API has been fixed.

See https://github.com/tsuru/postgres-api/pull/14 for details.

I have also added rake to Gemfile as it was missing for some reason.
# How To Test

To test, please make sure you run Ansible to have your environment
up to date (newest version of Postgres API) and run smoke tests:

```
make test-gce DEPLOY_ENV=yourenvironment
```
# Who Should Test This

Anybody except @RichardKnop.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ gem 'rspec'
 gem 'serverspec', '~> 2.18.0'
 gem 'ansible_spec'
 gem 'highline'
-
+gem 'rake'
 
 
 
--- a/spec/endtoend/endtoend_spec.rb
+++ b/spec/endtoend/endtoend_spec.rb
@@ -144,15 +144,7 @@ describe "TsuruEndToEnd" do
       expect(@workspace.tsuru_command.stdout).to include query
     end
 
-    it "should be able to connect to the applitation via HTTPS with a valid cert" do
-      pending "We don't have a certificate for this :)"
-      sampleapp_address = @workspace.tsuru_command.get_app_address(@sampleapp_name)
-      response = URI.parse("https://#{sampleapp_address}/").open()
-      expect(response.status).to eq(["200", "OK"])
-    end
-
     it "should be able to unbind and bind a service to an app" do
-      pending "There is already a bug filed: https://github.com/tsuru/postgres-api/issues/1"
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -3,7 +3,7 @@ gem 'rspec'
 gem 'serverspec', '~> 2.18.0'
 gem 'ansible_spec'
 gem 'highline'
-
+gem 'rake'
```
dcarley, line 6: Could you break this and the `Gemfile.lock` change into a separate commit which explains why we need to add this?

spec/endtoend/endtoend_spec.rb
```diff
@@ -144,15 +144,7 @@
       expect(@workspace.tsuru_command.stdout).to include query
     end
 
-    it "should be able to connect to the applitation via HTTPS with a valid cert" do
```
dcarley, line 147: Why did this move to the bottom? It should probably be next to the other HTTPS test, which it would replace, if anything.

RichardKnop, line 147: I see. We thought since it's the only test which  is being skipped right now, it should be at the end of the test suite. We can move it back.

Gemfile
RichardKnop, line 6: Done by Hector.

spec/endtoend/endtoend_spec.rb
keymon, line 147: Personally I like it like this, so the pending tests are at the end. Also, before it was not next to the HTTPS test, it had `should get log output in 3 seconds` before.

dcarley, line 147: np. I think that it too would have benefited from being a separate commit saying that, then I wouldn't have needed to ask. But not a blocker for merging.

dcarley:
Ansible pulled and deployed the new version of postgresapi automatically. Tests ran fine against my environment :+1:

spec/endtoend/endtoend_spec.rb
keymon, line 147: Thx for merging. 

True, should be in a different commit. Gone now :)

PR #145, opened 2015-07-29
[#99772862] deploy etcd unit

mtekel:
[99772862 deploy etcd unit](https://www.pivotaltracker.com/n/projects/1275640/stories/99772862) -  Add to etcd our deployment. This is needed for integrating vulcanD router. For the purpose of closest comparison with hipache we deploy it in similar configuration to redis: single node cluster running on tsuru-db server.
#### Testing

Run the tsuru_db playbook having set vulcand variable to true: 
`ansible-playbook -i gce.py -e "@platform-gce.yml" -e "deploy_env=myEnv" tsuru_db.yml -e "vulcand=true" -vvvv`

Once deployed, you can test etcd (from one of hosts within your environment):
- insert data by `curl -L http://<tsuru_db_internal_IP>:2379/v2/keys/message -XPUT -d value="Hello world"`
- retrieve data via `curl -L http://<tsuru_db_internal_IP>:2379/v2/keys/message`

You can also test performance using apache bench:
- writing: `ab -u <(echo value=\"bar\") -n 100000 -c 200 http://<tsuru_db_internal_IP>:2379/v2/keys/bench`
- reading: `ab -n 100000 -c 200 http://<tsuru_db_internal_IP>:2379/v2/keys/bench`

You should get around 5000 req/s when writing and when 6000 reading.
#### Reviewing

Anyone can review this except @keymon and @mtekel

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -10,6 +10,8 @@
   version: v1.2.2
 - src: Stouts.grafana
   version: 2.0.1
+- src: retr0h.etcd
+  version: 9a612f2ab2dfd52a9be329030bb2227c4d582423
 
 # GDS created and maintained playbooks
 - name: docker_server
--- a/site.yml
+++ b/site.yml
@@ -1,33 +1,6 @@
 # Bring up db tier (redis, mongodb).
 - include: registry.yml
-
-- hosts: "{{ hosts_prefix }}-tsuru-db"
-  sudo: yes
-  vars:
-    mongodb_conf_dbpath: "/var/lib/mongodb"
-  roles:
-    - bennojoy.redis
-    - greendayonfire.mongodb
-
-#FIXME: Remove when deployed to all environments
-  post_tasks:
-    - name: MongoDB | Check /data/db directory
-      stat: path=/data/db
-      register: mongo_data_db
-    - name: MongoDB | Migrate data - stop mongo
-      service: name=mongod state=stopped
-      when: mongo_data_db.stat.isdir is defined and mongo_data_db.stat.isdir
-    - name: MongoDB | Migrate data - prepare destination
-      shell: >
-        rm -rf {{ mongodb_conf_dbpath }}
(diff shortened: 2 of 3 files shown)


requirements.yml
```diff
@@ -10,6 +10,7 @@
   version: v1.2.2
 - src: Stouts.grafana
   version: 2.0.1
+- src: retr0h.etcd
```
dcarley, line 13: Could you pin the (latest) version of this so that we don't unwillingly upgrade?

mtekel, line 13: they have no release, added latest commit hash as version

dcarley, line 13: :+1:

RichardKnop:
:+1:

dcarley:
I tested this and it was fine. It doesn't bind to loopback so `etcdctl` doesn't work, but I don't think that matters. The only thing I wasn't sure about was that the cluster node was associated with the machine's IP address (`-initial-cluster 10.128.11.66=http://10.128.11.66:2380`). Does that matter? Presumably if we recreated the machine then we'd just create a new cluster and data (like MongoDB and Redis)?

PS: Sorry for the delay.

mtekel:
The cluster config is defined here: https://github.com/retr0h/ansible-etcd/blob/master/templates/etc/init/etcd.conf.j2
We'd have to fork the playbook if we need to change this. In GCE it should end up as `tsuru_db=http://tsuru.db.internal.ip:2380`. I am also not 100% sure about port numbers, but these are easily configurable via variables: https://github.com/retr0h/ansible-etcd/blob/master/defaults/main.yml

This is the initial configuration that we would possibly need to modify as we get experience with what VulcanD really needs/prefers.

dcarley:
> This is the initial configuration that we would possibly need to modify as we get experience with what VulcanD really needs/prefers.

Yeah, I'm not sure that it matters right now. Good to merge :+1:

PR #146, opened 2015-07-30
[#99420212] Install all Tsuru packages from our PPA mirror

dcarley:
Install the Tsuru packages for the core components Hipache, Gandalf, and
API, from our own PPA which is a mirror of the upstream PPA. This allows us
to pin the version that we use for stability.

A limitation of PPAs whereby they only present the latest version meant that
if we have pinned the version number in our code then we would have been
unable to install when Tsuru release a new version.

We have to use `pre_tasks` to remove the upstream PPA which is no longer
installed by the playbooks when the `tsuru_repo` var is passed. This can be
removed once it has been deployed to all existing machines, so long as we
continue to pass `tsuru_repo`.

The next time we upgrade the packages in our PPA we'll need to think about
how we roll it out to existing environments. Whether we change all of the
playbooks to `state=latest` (which is a pain to variabilise because it's a
separate attribute from the version which we currently expose) or pin the
specific version (which is also a pain because we'd have to time the changes
perfectly). It feels like it's out of scope for now though.

---

I have tested this on a new and existing environment in AWS.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -19,13 +19,13 @@
   version: v0.1.1
 - name: hipache
   src: https://github.com/alphagov/ansible-playbook-hipache.git
-  version: v0.0.2
+  version: v0.1.0
 - name: tsuru_api
   src: https://github.com/alphagov/ansible-playbook-tsuru_api.git
-  version: v0.0.4
+  version: v0.1.0
 - name: gandalf
   src: https://github.com/alphagov/ansible-playbook-gandalf.git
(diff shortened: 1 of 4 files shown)


keymon:
I think soon we will need to do the `state=latest`. For instance we need it to install the latest snapshot.

 If you pin the version it will use the pinned version, so it will be the same than `state=present`. If you don't pin the version, I understand that you want the latest version.

But yes, probably that is part of a different PR.

PR #149, opened 2015-07-30
[#100192442] Resuming AWS environment overwrites EC2 cache

mtekel:
In contrary to what was expected, ec2.py in ec2-wake was executed in the local directory, overwriting cache with stopped instances. This PR ensures that ec2.py executes in /tmp, where it will keep and handle its local cache.


wake-ec2.yml
```diff
@@ -1,12 +1,14 @@
...
-      - name: Deploy ec2.py
-        file: src=ec2.py dest=./ec2.py mode=0744
+      - name: Copy ec2.py to /tmp
+        copy: src="./ec2.py" dest="/tmp/ec2.py" mode=0744
+      - name: Copy ec2.ini to /tmp
+        copy: src="./ec2.ini" dest="/tmp/ec2.ini" mode=0644
       - name: Get stopped instances
         shell: "./ec2.py --stopped --refresh-cache"
+        args:
+          chdir: /tmp
```
dcarley: I'm not sure about hardcoding `/tmp` and it all seems a bit complicated (compared to GCE).

How about instead resetting the cache with `./ec2.py --refresh-cache` afterwards?

mtekel: I initially thought about it, but main reason I avoided that is to not slow things down anymore. On further tests, instances enter 'pending' state first, so they would not be listed anyway (we filter for only 'running' instances in ec2.ini). Instead, let's add option to invalidate cache to ec2.py.

mtekel:
It's a part of a story - comment and reaction to it in the form of a new commit. I could rebase it all...

keymon:
I rebased to master.

keymon:
:+1:  Works fine for me.

PR #150, opened 2015-07-31
[#99420212] Move tsuru_repo variables to tighter scope

dcarley:
Apply them to the role that's using them rather than leaking them
everywhere. I should have done this in the first place. The `vars` in these
files are abstractions that are used further down.

---
#147 should go first and I'll rebase this after.


dcarley:
I've rebased and added another commit to catch two other similar vars.

mtekel:
Works correctly (uses correct ppas), passes smoketests, merging.

PR #152, opened 2015-07-31
[#98108998] SPOF on Ansible Galaxy

mtekel:
Ansible galaxy server resolves playbook names into links where to download these playbooks from. In our case, all these playbooks are on github. Put the resolved values directly into requirements.yml (ansible galaxy prints the resolution results when you run it), so that we don't rely on ansible galaxy server anymore.


dcarley:
:+1: :+1: :+1:

PR #153, opened 2015-07-31
[#99772862] Trim vulcand etcd condition

mtekel:
Remove `and vulcand` from `when: vulcand is defined and vulcand`. We don't want to test for specific contents of vulcand variable, just whether it is defined or not. This way we can use the value of variable to carry some other useful information, like vulcand version, or some other specific setting.


dcarley:
Thanks, using this for my branch too.

PR #156, opened 2015-08-03
[#99773312] Enable vulcand

dcarley:
#### Rename var hipache_port to router_port

To reflect the fact that it will be used for both Hipache and Vulcand. We
can move it to `group_vars` at the same time, to match `docker_port` and
`redis_port`, which makes it slightly clearer that the variable name is
different from the one being passed to the actual role.
#### Optionally install vulcand on the router nodes

Instead of Hipache, when the `vulcand` feature flag is enabled. Version is
the latest stable. Port is the same as Hipache. It's using the single etcd
endpoint which is running on the DB host.

Per the comment in `requirements.yml`, I haven't pinned the version of the
playbook because it may be under active development while we're testing and
it'll be easier to not make two stage releases. We should version it if we
continue to use it in the future.

NB: This isn't intended to convert an existing environment from Hipache to
Vulcand. The ports of the routers will clash and the previously deployed applications won't be properly registered in vulcand. We only intend to create new environments, for testing, with this feature flag.
#### Optionally configure Tsuru API for vulcand

When the `vulcand` feature flag is enabled. The playbook version is being
bumped at the same time (with backwards incompat changes) so that we can
configure the router. I've arbitrarily chosen/set the API port, because we
need to pass the same value to both vulcand and Tsuru API.

I'm going to create tech debt stories for the following:
- the Tsuru API is only using the first router host for Vulcand's API. If we
  use vulcand beyond testing then we should fix this single point of
  failure
- the variable `hipache_host_external_lb` and the DNS hostname that is
  reference should be renamed from "hipache" to "router", so that it's
  agnostic of the actual router we use

---

In order to test this ~~you will need to workaround [missing queue configuration in `tsuru.conf`](https://www.pivotaltracker.com/story/show/100310408) (merged)~~. Additionally an unrelated [test for log output will fail](https://www.pivotaltracker.com/story/show/100407030) and we discovered a problem [removing applications that have CNAMEs](https://www.pivotaltracker.com/story/show/100406746).

### DIFF
--- a/group_vars/all/globals.yml
+++ b/group_vars/all/globals.yml
@@ -1,5 +1,9 @@
 ---
 
+etcd_host_name: "{{ hosts_prefix }}-tsuru-db"
+etcd_host: "{{ hostvars[groups[etcd_host_name][0]][ip_field_name] }}"
+etcd_port: 2379
+
 redis_host_name: "{{ hosts_prefix }}-tsuru-db"
 redis_host: "{{ hostvars[groups[redis_host_name][0]][ip_field_name] }}"
 redis_port: 6379
--- a/group_vars/all/globals.yml
+++ b/group_vars/all/globals.yml
@@ -18,6 +18,9 @@ docker_port: 4243
 docker_version: 1.7.0
 
 router_port: 8080
+router_api_port: 8081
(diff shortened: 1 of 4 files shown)


actionjack:
:+1:

PR #163, opened 2015-08-06
[#99119628] wake-gce: wake instances in parallel

mtekel:
GCE has been recently getting slower and waking all instances one by one is taking unexpectedly long. Worst case scenario is that all requests time out after 180 seconds. There's no need to start instances one by one. Wake them all in parallel. The children of ansible run (gce.py script instances waking your instances up) remain running till they succeed or time out. You can then check manually what's the state of your environment and if appropriate run the wake script again, which will attempt to start all instances in "terminated" state.

### DIFF
--- a/wake-gce.yml
+++ b/wake-gce.yml
@@ -1,15 +1,8 @@
 ---
-
 # Start all the instances in the specified GCE deployment
   - hosts: localhost
     tasks:
-      - name: Deploy gce.py
-        file: src=gce.py dest=./gce.py mode=0744
-      - name: Deploy secrets.py
-        file: src=secrets.py dest=./secrets.py mode=0700
-
       - name: Wake instances
         shell: "./gce.py --start {{ item }} &"
         when: item | match("^{{ deploy_env }}-*")
         with_items: groups['status_terminated']
-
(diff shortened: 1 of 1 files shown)


wake-gce.yml
```diff
@@ -9,7 +9,7 @@
         file: src=secrets.py dest=./secrets.py mode=0700
 
       - name: Wake instances
-        shell: "./gce.py --start {{ item }}"
+        shell: "./gce.py --start {{ item }} &"
```
dcarley, line 6: Would it be better to use [`async`](http://docs.ansible.com/ansible/playbooks_async.html) so that Ansible doesn't exit until they are all finished?

mtekel, line 6: Maybe, I have not tried and this has created desired result...

mtekel:
To be done in a nicer way.

dcarley:
For reference, we talked about this and weren't sure whether backgrounding the jobs would make it harder to know when and debug if they failed. Ansible doesn't support `async` and `with_items` together until Ansible 2.0. It's non-trivial to construct a list and pass to `xargs -P`.

PR #164, opened 2015-08-07
[#91207216] suspend instances via API

mtekel:
Before this PR we used ansible to suspend our instances simply by using `-a 'sudo shutdown'` option. This is no longer possible, because coreOS we have introduced (and plan upgrading to) doesn't really support running ansible (or anything else) outside containers. And ansible doesn't support running inside containers (even though in version 3.0 it might). We have thus decided to suspend instances via cloud APIs, like we do wake them up. Note that this is safe, because cloud providers send AHCI power off signal to the VMs (equivalent to you pressing power button on your atx-compliant computer). The VMs then shut down normally and end up in `stopped` (on AWS) or `terminated` (on GCE, which means stopped) state.

In this PR we:
- extend functionality of gce.py script to suspend instances
- adjust ec2 and gce wake playbooks to also support suspending
- adjust makefile for the changes made in playbooks

Who can review:
Anyone but @mtekel


dhilton:
Created a new GCE environment, checked out this branch and run suspend. The GCE console took a while to replicate the change but all of those machines were suspended.

dhilton:
LGTM :+1:

PR #165, opened 2015-08-11
[#91207216] Configure coreOS docker nodes

mtekel:
https://www.pivotaltracker.com/story/show/91207216 
Depends on https://github.com/alphagov/tsuru-terraform/pull/96

**What**

Integrate coreOS nodes deployed in Depends on https://github.com/alphagov/tsuru-terraform/pull/96 to the tsuru platform by registering them as docker nodes. Also modify ansible configuration to not try to deploy telegraf on these nodes, as in coreOS nothing runs directly on host, but in container and there is a different way of monitoring these nodes (see various [options](https://www.pivotaltracker.com/n/projects/1275640/stories/100653964) in the comments).

**How to test**

Do terraform deploy of https://github.com/alphagov/tsuru-terraform/pull/96. Then run ansible configuration from this branch. All should complete successfully. After done configuring, run smoketests to verify platform works correctly. Note that currently smoketests are broken and are waiting for https://github.com/alphagov/flask-sqlalchemy-postgres-heroku-example/pull/6 to be merged to fix that.

 **Who should review**

Anyone but @mtekel or @dhilton


dcarley:
NB: the `Register coreos docker nodes` task produces a lot of output, because it logs a `skipping` entry for every VM in the provider (including those for other environments). Michal and I couldn't find a way around this for the moment and it doesn't seem to make the operation slower, so we're leaving for now.

PR #169, opened 2015-08-21
[#101503740] Docker nodes not deleted if reprovisioned

mtekel:
https://www.pivotaltracker.com/n/projects/1275640/stories/101503740

Currently docker nodes are coreOS based. CoreOS is configured via cloud-config, which is stored in VM user-metadata and applied on the instance during the startup. However, user-metadata is immutable in AWS, so every time you change it, terraform re-creates new instances. Ansible run then registers these new instances in tsuru, but we were never removing the old ones. This places burden on tsuru, as it still tries to reach and healthcheck the old instances, which don't exist anymore.

This PR adds cleanup step, which will only leave nodes currently existing registered in tsuru and remove all others.

**Testing**

Apply. Add some non-existing node to a pool, e.g. `tsuru-admin docker-node-add address=http://1.2.3.4:4567 pool=default --register`. Run tsuru_api.yml playbook and watch this node being removed.

**Reviewing**
Anyone but @mtekel or @saliceti

### DIFF
--- a/tsuru_api.yml
+++ b/tsuru_api.yml
@@ -8,9 +8,6 @@
     - name: Tsuru Api | Remove our private repo
       apt_repository: repo='ppa:multicloudpaas/tsuru' state=absent
       when: vulcand is defined
-      # FIXME: Remove when `tsuru_repo` is deployed.
-    - name: Tsuru Api | Remove upstream APT repo
-      apt_repository: repo='ppa:tsuru/ppa' state=absent
   vars:
     router_config_vulcand:
       domain: "{{ hipache_host_external_lb }}"
--- a/tsuru_api.yml
+++ b/tsuru_api.yml
@@ -2,14 +2,14 @@
   sudo: yes
   pre_tasks:
     - include: ssl_proxy_pre.yml
-    - name: Remove daily snapshots repo
+    - name: Tsuru Api | Remove daily snapshots repo
       apt_repository: repo='ppa:tsuru/snapshots' state=absent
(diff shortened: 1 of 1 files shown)


dcarley:
Nice approach, as well as good atomic and well described commits :+1: 

> Add some non-existing node to a pool, e.g. tsuru-admin docker-node-add address=http://1.2.3.4:4567 pool=default --register. Run tsuru_api.yml playbook and watch this node being removed.

I wasn't able to get this to work, it gave me:

```
TASK: [Tsuru Api | Remove zombie nodes from the default pool] *****************
failed: [10.128.11.146] => (item=1.2.3.4) => {"changed": true, "cmd": "tsuru-admin docker-node-remove http://1.2.3.4:4243 -y", "delta": "0:00:00.011991", "end": "2015-08-24 12:18:45
.464181", "item": "1.2.3.4", "rc": 1, "start": "2015-08-24 12:18:45.452190", "warnings": []}
stderr: Error: node with address "http://1.2.3.4:4243" not found in cluster
skipping: [10.128.11.146] => (item=10.128.11.248)
skipping: [10.128.11.146] => (item=10.128.11.83)
skipping: [10.128.11.146] => (item=10.128.13.192)
```

I think I've solved this in fc2f543b915c902673876005ed4aa433b5346928. If you're happy with that approach, could you cherry-pick it to this pull request and then this is good to merge.

mtekel:
Updated the branch with your commit....

dcarley:
Thanks!

PR #172, opened 2015-11-03
[#107185318] Reconfigure the nginx on Tsuru to allow larger file upload

Jonty:
Rather than setting this to 0 and allowing uploads of any size I limited it to 512m, as anything larger than this is really getting silly for deployment on a PaaS.


combor:
:+1:

PR #174, opened 2015-11-09
[#106191568] Tsuru: Update components for metrics collection and tag instances

combor:
[#106191568 Monitoring dashboard for the trial Tsuru platform](https://www.pivotaltracker.com/story/show/106191568)
# What

We want to build a dashboard to monitor tsuru using grafana. To do so we need to collect metrics from all Tsuru servers with the right tagging.
# Context.

To enable metrics collection we had to:
- Upgrade Telegraf to version 0.2.0
- Upgrade influxdb to version 0.9.4.2
- Upgrade requirements file
- Added tags for servers for they `type` and the `instance_name`, so we can aggregate metrics for specific type of hosts ( api, postgres, ...).

>  Note: As we lack of roles in our ansible config, we do rely on the names of the instances of the VMs, using a regular expression to extract the type. To improve it, the alternative would be refactor how terraform+ansible work to introduce ansible roles, [as described here](https://www.pivotaltracker.com/story/show/100111294)
- Fix grafana db creation as we found [api breaking change](https://github.com/grafana/grafana/blob/master/CHANGELOG.md#210-2015-08-04) for v 2.1
- Fix influx DB password
# How to test it
1. Start the environment both AWS and GCE
2. go to http://<envname>-grafana.tsuru.paas.alphagov.co.uk:3000 or http://<envname>-grafana.tsuru2.paas.alphagov.co.uk:3000
3. Create a graph and confirm that metrics from CoreOS with the right tags are arriving. (you can import this dashboard if you want: https://gist.github.com/keymon/17d193b1a7b6b61fc6a3)
4. Check if influx-db has the right name and is running.
# Who can review this?

Anyone but @colin, @keymon or @combor


mtekel:
Mostly works. Depends on firewall rules adjustment to enable NAT and Gandalf to send metrics to Influx as well.

keymon:
@mtekel We'll fix the firewall rules in a different PR

mtekel:
This works, I can see metrics from all hosts.

PR #178, opened 2015-11-11
[#106191568] Bump influxdb role to 0.0.4

saliceti:
# What

Fix in influxdb role to allow automatic restart when the package is upgraded.
# How to review
- Review https://github.com/alphagov/ansible-playbook-influxdb/pull/3 should be sufficient
# Who can review

Anyone but @saliceti or @keymon


mtekel:
Works as described. There is influxDB service restart every time you:
- install
- upgrade version
- downgrade version

